PR #6, opened 2023-12-30
Fix tg parser

aleksspevak:


ohld (approved):

ohld:
now sure that this will work, I'll fix that rn if fails on staging

PR #10, opened 2024-01-18
Updated to v4 API

zakhar-kogan:
1. File loader updated + URL concatenation added
2. Pipeline runner updated for v4 - Request structure changed - URL changed


ohld:
Please provide a script here how you tested that this fix actually works.

PR #11, opened 2024-01-22
caption under meme

VeryBigSad:


src/tgbot/senders/next_message.py
```diff
@@ -36,17 +35,24 @@ async def next_message(
...
+
+    send_new_message = (
+        prev_reaction_id is None or Reaction(prev_reaction_id).is_positive
+    )
     if not send_new_message and prev_update_can_be_edited_with_media(prev_update):
         msg = await prev_update.callback_query.message.edit_media(
             media=get_input_media(meme),
             reply_markup=meme_reaction_keyboard(meme.id),
         )
+        meme_caption_with_referral_link = (f"{meme.caption}\n\n" if meme.caption else "") + f'<a href="https://t.me/ffmemesbot?start=s_{user_id}_{meme.id}">FastFoodMemes</a>'
```
ohld, line 47: 1. move function to utils.py
2. fix code line length codestyle
3. hardcoded "ffmemesbot" is not ok for local / staging / production environment testing
4. is "s_" a good option?
5. have you tested that this deeplink value appears in users table? What if a user was already created?

VeryBigSad: 1. done
2. done
3. Put that into `.env`
4. `s_` had been used in legacy ffmemesbot code, stands for "share".
5. In case if we see `/start` from a created user, we update only user-changable fields. Like username, first/last name, is_premium, etc.

src/tgbot/senders/meme.py
```diff
@@ -40,27 +41,60 @@ async def send_new_message_with_meme(
...
+
+async def edit_last_message_with_meme(
+    user_id: int,
+    meme_id: int,
+    meme: MemeData,
+):
+    bot_username = bot.application.bot.username
+    meme_caption_with_referral_link = (
+        (f"{meme.caption}\n\n" if meme.caption else "")
+        + f'<a href="https://t.me/{bot_username}?start=s_{user_id}_{meme.id}">FastFoodMemes</a>'
```
ohld, line 86: вынести в функцию же

src/tgbot/service.py
```diff
@@ -21,13 +18,20 @@ async def save_tg_user(
     id: int,
     **kwargs,
 ) -> None:
+    update_dict_if_user_already_created = {
```
ohld, line 21: зачем это? не надо

ohld (changes requested):

src/tgbot/senders/meme.py
ohld: да и кодстайл. Ад какой-то

src/tgbot/senders/next_message.py
ohld: 3.  `bot.application.bot.username` - does it really stores bot's username? Cool finding if that's true

ohld (approved):

PR #22, opened 2024-02-06
Faster reaction times

VeryBigSad:
Do database insert, send new meme & redis update all at the same time.
Since in the recommendation function we check the existence of meme-reaction pair, not it having an actual reaction, we don't need to do it before deciding on the next meme.
We also don't need to check whether we've reacted to that meme in the past, because if we did - the buttons would automatically disappear via the webhook response.

### DIFF
--- a/src/tgbot/handlers/reaction.py
+++ b/src/tgbot/handlers/reaction.py
@@ -25,7 +25,7 @@ async def handle_reaction(update: Update, context: ContextTypes.DEFAULT_TYPE) ->
     )
 
     await asyncio.gather(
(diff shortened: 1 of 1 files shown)


ohld:
> We also don't need to check whether we've reacted to that meme in the past, because if we did - the buttons would automatically disappear via the webhook response.

Have you tested this? If a user click on reaction button fast, this may result in multiple memes send in return.

ohld:
Also you've mentioned that this has faster reaction times. Do you have any proofs of that? What was the reaction time before and what is it it your version?

PR #31, opened 2024-02-20
fix sending non-image memes to Mystic OCR

VeryBigSad:
Currently, we are sometimes sending files that are not images to Mystic OCR, resulting in errors and more bandwitdth being used.
This PR fixes it.

### DIFF
--- a/src/storage/service.py
+++ b/src/storage/service.py
@@ -228,7 +228,8 @@ async def get_memes_to_ocr(limit=100):
 
 
 async def get_unloaded_tg_memes(limit) -> list[dict[str, Any]]:
-    """Returns only MemeType.IMAGE memes"""
+    """Returns memes from Telegram, that have not been yet uploaded to Telegram."""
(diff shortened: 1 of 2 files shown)


src/storage/service.py
```diff
@@ -254,11 +255,12 @@ async def get_unloaded_tg_memes(limit) -> list[dict[str, Any]]:
...
 
 async def get_unloaded_vk_memes() -> list[dict[str, Any]]:
-    "Returns only MemeType.IMAGE memes"
+    """Returns memes from VK, that have not been yet uploaded to Telegram."""
+
     select_query = f"""
         SELECT
             meme.id,
-            '{MemeType.IMAGE}' AS type,
+            meme.type,
```
VeryBigSad, line 263: @ohld Not sure whether I'm fixing a bug here or this is intentional, need your feedback

ohld (approved):

PR #88, opened 2024-10-08
blender rollout; changed configs a bit

ledovsky:


ledovsky:
Changes:

- blender is rolled out
- reco batches are increased from 5 to 7 items
- lr_smoothed is only guaranteed at the first position

ohld (approved):

PR #137, opened 2026-03-28
feat(qa): E2E smoke tests for post-deploy verification

ohld:
## Summary
- Standalone Telethon script (`scripts/e2e_smoke.py`) that tests @ffmemesbot as a real Telegram user
- Two tests: `/start` → meme arrives, click like → next meme arrives
- QA Agent runs this post-deploy alongside existing canary/Sentry/DB checks
- Session string for +998997053933 generated on server, stored in Paperclip secrets

## Files
- `scripts/e2e_smoke.py` — smoke test runner (exit 0/1)
- `scripts/generate_session_string.py` — one-time session string generator
- `requirements-e2e.txt` — telethon dependency
- `agents/qa/AGENTS.md` — E2E instructions for QA agent
- `agents/.paperclip.yaml` — Telegram env vars as optional secrets

## Test plan
- [x] Tested locally: PASS (both tests)
- [x] Tested from server (t.ffmemes.com): PASS (both tests)
- [x] Paperclip secrets created (TELEGRAM_API_ID, TELEGRAM_API_HASH, TELEGRAM_SESSION_STRING)
- [ ] Verify QA agent picks up E2E tests on next post-deploy trigger

PR #138, opened 2026-03-28
feat(infra): wire Paperclip autonomous loop for April vacation

ohld:
## Summary
- Fix 3 dead Paperclip routines by adding cron triggers (Analyst daily, CEO weekly, gstack update)
- Add secured webhook proxy at `/webhooks/qa-alert` for Sentry/Coolify → Paperclip QA alerts
- Add process health watchdog routine (daily, checks all routines are running)
- Enforce gstack sprint workflow in all agent AGENTS.md files
- Prefect flow failures now notify Paperclip QA agent

## What's already live in Paperclip (API calls, no deploy needed)
- 3 cron triggers on previously-dead routines
- Watchdog routine created with daily cron
- CEO /retro skill + Comms /frontend-design skill attached
- New webhook trigger for proxy (bearer auth)
- 4 prod env vars set via Coolify (PAPERCLIP_QA_TRIGGER_URL/SECRET, WEBHOOK_PROXY_SECRET, SENTRY_CLIENT_SECRET)

## Test plan
- [ ] CI passes (lint + tests)
- [ ] Staff Engineer reviews via Paperclip trigger (this PR should auto-trigger it)
- [ ] After merge: Coolify auto-deploys
- [ ] After deploy: verify `/healthcheck` responds
- [ ] After deploy: verify `/webhooks/qa-alert` returns 403 without auth
- [ ] After deploy: Coolify webhook fires QA agent (if deploy notification configured)
- [ ] Next cron cycle: Analyst daily report fires at 06:00 UTC
- [ ] Next cron cycle: Watchdog fires at 12:00 UTC


ohld:
## Staff Engineer Review — PR #138

**Verdict: Needs fix before merge.** 1 real bug + 4 informational issues.

---

### 🔴 Bug: Paperclip QA notification silently skipped when Telegram is unconfigured

**`src/flows/hooks.py:20-24`** — `notify_telegram_on_failure` returns early if `TELEGRAM_BOT_TOKEN` or `ADMIN_LOGS_CHAT_ID` is not set. The new `notify_qa_sync(...)` block is appended at the end of the function body (lines 49–54), which means **it is never reached** when Telegram creds are missing.

This defeats the feature: Paperclip can't receive flow failure alerts independently of Telegram.

**Fix:** Move `error_msg` extraction and `notify_qa_sync` call before the early return, so Paperclip gets notified regardless of Telegram config.

---

### ⚠️ Informational (no block)

**1. `paperclip.py:82-90` — Webhook secret in query string appears in access logs.** `?secret=...` is logged by every HTTP server/proxy. Low risk for internal webhook, but consider `X-Webhook-Secret` header instead.

**2. `paperclip.py:107-115` — Non-Sentry callers can spoof `Sentry-Hook-Resource`.** Only trust that header when `is_sentry is True`.

**3. `paperclip.py:124-133` — Proxy always returns 200 even on downstream failure.** Intentional (prevents Sentry retries), but silent. The status code is logged — just make sure someone is watching.

**4. `hooks.py:49` — `notify_qa_sync` is a synchronous blocking call (up to 10s) on the flow failure path.** Fine at current scale; revisit if Prefect worker thread exhaustion shows up during outages.

---

**CI:** GitHub Actions ✅

The blocking issue is the Telegram gate bug — a 10-line fix in `hooks.py`. Everything else is solid.

ohld:
**Staff Engineer Review: APPROVED**

No critical issues. One informational finding (not a blocker).

## Findings

**INFORMATIONAL — Test coverage gap**
`src/integrations/paperclip.py` has no tests for the auth functions (`_verify_sentry_signature`, `_verify_shared_secret`). The HMAC logic and constant-time comparisons are correct, but these paths are worth unit testing since they gate the security boundary. Not a blocker for this PR — worth tracking as a follow-up.

## What's correct

- HMAC-SHA256 verification with `hmac.compare_digest` (constant-time) — correct
- Shared secret comparison with `hmac.compare_digest` — correct
- Coolify UUID filter (`FFMEMES_APP_UUID`) prevents forwarding events for unrelated apps
- Graceful degradation when `PAPERCLIP_QA_TRIGGER_URL` / `PAPERCLIP_QA_TRIGGER_SECRET` not configured
- `notify_qa_sync` uses synchronous `httpx.post` (correct for Prefect hook context)
- `webhook_proxy_qa` uses `async httpx.AsyncClient` (correct for FastAPI)
- Conflict resolution: kept both E2E smoke tests section (from PR #137) and Process Health Check watchdog — both sets of secrets in `.paperclip.yaml`
- All CI checks pass

Merging.

PR #140, opened 2026-03-28
fix(cold-start): lower cold_start_explore reaction threshold 50→20 [FFM-66]

ohld:
## Summary

- **Root cause**: `cold_start_explore` Phase 1 was returning 0 candidates for all new users (Mar 27–28), causing fallback to `lr_smoothed` instead.
- **Why**: The `>= 50 (nlikes+ndislikes)` threshold was miscalibrated. `nlikes+ndislikes` counts only explicit like/dislike reactions, not skips. At ~30–40% explicit-reaction rates, a meme needs ~150 sends to accumulate 50 reactions — far too strict.
- **Fix**: Lowered to `>= 20` reactions. Quality gate `lr_smoothed >= 0.40` unchanged. Consistent with `GOAT_MIN_REACTIONS = 10` ("enough statistical signal").

## Impact

68+ new users on Mar 27 and subsequent users received `lr_smoothed` fallback instead of the intended `cold_start_explore`. The excellent observed metrics (39.7% first-meme LR, 57% 10-meme retention) are from `lr_smoothed` quality, not cold_start_v2 Phase 1. This fix restores the experiment to its intended code path.

## Test Plan

- [ ] Deploy and confirm `cold_start_explore` engine_stats show non-zero meme-sends for next new user batch
- [ ] Verify next analyst report shows `cold_start_explore` active (meme_number 1-5 for new users)
- [ ] No regression in first-meme LR or 10-meme retention

Refs: FFM-66

Generated with Claude Code


ohld (commented):
## Staff Engineer Review

**Pre-Landing Review: 0 critical, 2 informational, 1 note on CI**

Code is clean. All changes are well-scoped and fix real user-facing bugs.

### CI Status

`lint` check fails — but this is **pre-existing on `production`**. The `I001` import ordering issues in `scripts/e2e_smoke.py` and `scripts/generate_session_string.py` exist on `production` too. This PR does not introduce new lint failures. Tests pass ✓

### Findings

1. **[INFORMATIONAL]** `src/tgbot/senders/meme.py:151` — `Forbidden` raised by `_do_send(None)` (the HTML-entity retry) will not be caught by the outer `except Forbidden` clause. Python sibling `except` clauses do not cover exceptions raised within each other's handlers. If a user blocks the bot simultaneously with an HTML entity meme error, `update_user(UserType.BLOCKED_BOT)` won't fire. Scenario is extremely unlikely in practice but is technically a regression from previous behavior. Fix: wrap the retry call in its own try/except Forbidden.

2. **[INFORMATIONAL]** `tests/tgbot/test_agent_tools.py:76` — Assertion is too permissive: `assert "error" not in result.lower() or "not found" in result.lower() or "sent" in result.lower()` passes for almost any string due to the `or` chain. Tighten to assert a specific success case.

### Core Changes

All three bug fixes are correct:
- Cold-start threshold 50→20: well-reasoned, well-documented. The >50 explicit-reaction requirement was unreachable for most memes (requires ~150 sends at 30-40% reaction rate). 68+ users affected.
- `execute_send_meme` string guard: clean fix for Sentry FF-BACKEND-VV.
- HTML entity retry: correct refactor, retry without parse_mode is the right approach.
- Webhook proxy silent failure fix: `resp.raise_for_status()` in `notify_qa_sync` is correct.

---

*Not merging due to CI lint failure (even though pre-existing). @ohld can merge manually if you're OK with the lint being pre-existing.*

PR #142, opened 2026-03-29
refactor(chat-agent): migrate to Agents SDK, fix DSML leaks

ohld:
## Summary
- Replace hand-rolled 177-line tool loop with `openai-agents` SDK `Runner.run()`
- Fix DSML XML leaks: DeepSeek sometimes outputs internal XML as plain text — added regex cleanup as safety net
- Add OpenAI tracing export for observability (traces → OpenAI dashboard)
- Use per-agent DeepSeek client (no global `set_default_openai_client` — avoids breaking other OpenAI calls)
- Drop broken `react_to_message` tool (model never received message_ids in context)
- Soften prompt: send memes when appropriate, not forced on every message
- Bump `openai` 1.75→2.8, add `openai-agents==0.4.*`

Net -113 lines. Same tool behavior (search, send, history), cleaner architecture.

## Test plan
- [ ] Verify `pip install` succeeds with new deps (openai 2.8 + openai-agents 0.4)
- [ ] Tag bot in group chat → get text response (no DSML XML garbage)
- [ ] Tag bot asking for memes → search_memes + send_meme tools fire correctly
- [ ] Verify existing OpenAI calls (ai.py vision, wrapped.py) still work with openai 2.8
- [ ] Check OpenAI dashboard → traces appear from chat agent
- [ ] Check `chat_agent_usage` table → rows logged with correct token counts


ohld:
## Staff Engineer Review — [FFM-96](/FFM/issues/FFM-96)

**Summary:** 3 issues found (0 critical, 3 informational). CI passes. Two bugs need fixing before merge.

---

### [BUG] `agents/backup/restore.sh:52` — Wrong key, restore always crashes

`restore.sh` reads `backup['agents_full']` but `backup.sh` writes the key as `'agents'`. Every restore attempt will throw `KeyError: agents_full`.

```python
# restore.sh:52 — fix this
for agent in backup["agents"]:   # was: backup["agents_full"]
```

---

### [SECURITY] `agents/backup/backup.sh` — Temp files world-readable + no cleanup on failure

Agent configs are written to `$TMP/agents.json` etc. On shared infra, `/tmp` is world-readable. If the script fails mid-run (network error, disk full), credential files stay on disk.

```bash
# After `TMP=$(mktemp -d)`, add:
chmod 700 "$TMP"
trap 'rm -rf "$TMP"' EXIT INT TERM
```

---

### [INFORMATIONAL] `agents/backup/restore.sh` — No guard on empty API responses

`curl -sf` returns exit 0 on HTTP 200 with an error body. If the API returns `{"error": "..."}`, `backup["agents"]` will be whatever that parsed to. Add a check that temp files are non-empty JSON arrays before running the Python loop.

---

**Non-blocking:**
- Channel ID fix (`-1001152876229` → `-1001472939243` for `@ffmemes`) is correct — important fix.
- Content policy in `comms/AGENTS.md` is well-scoped.
- README table matches live Paperclip state.

ohld:
## Pre-Landing Review — Staff Engineer

**Result: CLEAN** (1 bug fixed inline, 2 informational notes)

### Fixed (inline)

**runner.py:113** — Token counting TypeError: `raw.usage.prompt_tokens` can be `None` for intermediate tool-call turns on some providers (DeepSeek returns `None` on tool-call turns for some model versions). Without `or 0`, the TypeError propagates out of `run_chat_agent`, gets caught by chat.py's outer `try` block, and the user's response is silently dropped even though the LLM produced it. Fixed with `or 0` guard in commit 8481b48. CI passes.

### Informational (non-blocking, can be followup)

**runner.py:37-38** — `set_tracing_export_api_key(settings.OPENAI_API_KEY)` sends all chat agent conversations (user Telegram messages, tool calls) to OpenAI for tracing. Intentional per PR description. Worth adding an explicit `CHAT_AGENT_TRACING_ENABLED` flag so it doesn't silently activate whenever `OPENAI_API_KEY` is present for unrelated purposes (ai.py vision).

**runner.py:22 / chat.py:155** — `clean_response()` is called twice (inside `run_chat_agent` + in `handle_agent_trigger`). Currently idempotent. Also: DSML regex doesn't handle unclosed tags (truncated response) — opener leaks through verbatim. Both low risk.

### Verified clean

- SQL: all parameterized
- `meme_id` int coercion: already guarded with try/except
- `api_key="dummy"`: safe — `run_chat_agent` guards `if not settings.DEEPSEEK_API_KEY: return None`  
- Module-level tracing init: properly gated by `if settings.OPENAI_API_KEY`
- Context isolation via `ChatAgentContext`: correct

Net -113 lines. Good migration.

PR #148, opened 2026-04-01
feat(wrapped): localize + broadcast script

ohld:
## Summary
- Localize wrapped UI for non-Russian users (EN channel gate, all slide text, stats, buttons)
- Add broadcast script to notify ~1,134 qualified users about /wrapped
- Fix missing await in broadcast_text.py

## Test plan
- [ ] Deploy to production
- [ ] Run `docker compose exec app python scripts/broadcast_wrapped.py --dry-run` to verify counts
- [ ] Run broadcast

### DIFF
--- a/scripts/broadcast_wrapped.py
+++ b/scripts/broadcast_wrapped.py
@@ -1,0 +1,111 @@
+"""
+Broadcast wrapped notification to all qualified users.
+
+Run inside the app container:
+    docker compose exec app python scripts/broadcast_wrapped.py
+
+Or with --dry-run to preview without sending:
+    docker compose exec app python scripts/broadcast_wrapped.py --dry-run
+"""
+
+import asyncio
+import sys
+
+from sqlalchemy import text
+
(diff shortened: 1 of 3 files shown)


ohld:
**Staff Engineer Review**

## AUTO-FIXED (committed in 5e805ba)
- `wrapped.py:747` — `_next_btn("wrapped_5")` missing `ru` param in oneliner text fallback. EN users would see "Дальше →" instead of "Next →". Fixed.

## INFORMATIONAL
- `broadcast_wrapped.py:86-100` — Error classification uses string matching (`"blocked" in err`) instead of `telegram.error.Forbidden`. Low risk for a one-off script, but a transient error with "not found" in the message would permanently mark an active user as `blocked_bot`. Recommend switching to typed exceptions before running.

## Summary
CI passes. One bug auto-fixed. Localization is thorough across all slide paths. `await asyncio.sleep` fix in broadcast_text.py is correct. Ready to merge.

PR #154, opened 2026-04-01
fix(db): handle ConnectionDoesNotExistError with retry + pool invalidation

ohld:
## Summary

- `pool_pre_ping=True` catches most stale connections, but there's a race: Postgres can close a connection between the ping and the actual query
- When this happens asyncpg raises `ConnectionDoesNotExistError`, which was surfacing as unhandled `DBAPIError` in `fetch_all` (Sentry 7347427630) and the webhook handler (Sentry 7379710797)

**Two-layer fix in `src/database.py`:**
- `handle_error` event listener marks `ConnectionDoesNotExistError` as a pool disconnect — SQLAlchemy invalidates that connection and never returns it from the pool again
- Single retry in `fetch_one` / `fetch_all` / `execute` — if the race fires, the retry acquires a fresh healthy connection and the request succeeds transparently

## Test plan
- [ ] Deploy to production and monitor Sentry: 7347427630 and 7379710797 should stop producing new events
- [ ] Verify app logs show no `ConnectionDoesNotExistError` after deploy
- [ ] `docker ps` + `docker logs` health check per CLAUDE.md checklist

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -11,7 +11,7 @@ class Config(BaseSettings):
     DATABASE_POOL_SIZE: int = 20
     DATABASE_POOL_TTL: int = 60 * 20  # 20 minutes
     DATABASE_POOL_PRE_PING: bool = True
-    DATABASE_POOL_TIMEOUT: int = 5  # seconds to wait for a connection before raising; keep low to fail fast
+    DATABASE_POOL_TIMEOUT: int = 5  # seconds; keep low to fail fast on pool exhaustion
 
     REDIS_URL: RedisDsn
     REDIS_HEALTH_CHECK_INTERVAL: int = 30
--- a/src/database.py
+++ b/src/database.py
@@ -20,10 +20,12 @@ from sqlalchemy import (
     Table,
     UniqueConstraint,
     Update,
+    event,
     func,
     text,
 )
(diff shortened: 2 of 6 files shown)


ohld:
**Staff Engineer Review — APPROVED**

Code review complete. Two-layer fix is correct:

1. `handle_error` listener marks `ConnectionDoesNotExistError` connections as disconnected — pool invalidates them immediately
2. Single retry in `fetch_one`/`fetch_all`/`execute` handles the race where a stale connection slips past `pool_pre_ping`

**On write retry safety:** `ConnectionDoesNotExistError` fires in asyncpg's `_check_open()` BEFORE the query is sent to Postgres. No double-write risk — the query never executes on the first attempt.

**Pre-existing lint + conflicts fixed:** 3 E501 violations in unrelated files, ruff formatting in `describe_memes.py`, and a merge conflict in `wrapped.py` with the production hotfix. All resolved and CI now passes (lint ✓ + tests ✓).

**Minor informational (non-blocking):**
- `_is_stale_connection_error` uses `exc.__cause__` — `DBAPIError.orig` is the documented SQLAlchemy API but both work in practice
- `type().__name__` matching is slightly fragile; `isinstance()` against imported asyncpg class would be more robust

Merging.

PR #158, opened 2026-04-06
fix(db): retry on asyncpg InternalClientError

ohld:
## Summary

- Handles `InternalClientError: cannot switch to state 15; another operation is in progress` from asyncpg — a transient race where the pool hands out a connection whose previous async commit hasn't fully completed
- Marks `InternalClientError` connections as disconnect so the pool evicts them (same as existing `ConnectionDoesNotExistError` handling)
- Adds retry-once logic in `fetch_one`, `fetch_all`, and `execute` for this error class
- Sentry issue 7343447228 (8 occurrences, culprit: `/tgbot/webhook`)

## Test plan

- [ ] Verify ruff passes (pre-commit hook ran successfully)
- [ ] Deploy and monitor Sentry for recurrence of the `InternalClientError`
- [ ] Confirm no regressions in webhook response times under normal traffic

### DIFF
--- a/src/database.py
+++ b/src/database.py
@@ -63,17 +63,26 @@ engine = create_async_engine(DATABASE_URL, **_engine_kwargs)
 
 
 @event.listens_for(engine.sync_engine, "handle_error")
-def _mark_stale_connection_as_disconnect(context: object) -> None:
-    """Tell SQLAlchemy to invalidate pool connections that asyncpg reports as closed.
+def _mark_bad_connection_as_disconnect(context: object) -> None:
+    """Tell SQLAlchemy to invalidate pool connections that asyncpg reports as broken.
 
     pool_pre_ping catches most stale connections, but a race is possible: the server
     closes the connection between the ping and the actual query.  When that happens
     asyncpg raises ConnectionDoesNotExistError (a subclass of InterfaceError).
-    Marking it as a disconnect causes the pool to discard that connection so it is
-    never handed out again.
+
+    Similarly, under high concurrency the pool can hand out a connection whose
+    previous async operation hasn't fully completed.  asyncpg raises
+    InternalClientError ("cannot switch to state …; another operation is in
+    progress") in this case.
+
+    Marking either as a disconnect causes the pool to discard that connection so
+    it is never handed out again.
     """
     original = getattr(context, "original_exception", None)
-    if original is not None and type(original).__name__ == "ConnectionDoesNotExistError":
+    if original is not None and type(original).__name__ in (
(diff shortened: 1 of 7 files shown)


ohld:
## Pre-Landing Review: 0 critical, 4 informational

**Scope Check: CLEAN** — 4 commits, each well-scoped to the stated intent (Sentry 7343447228 + related stability fixes).

### CI Status

- **Lint FAIL**: `src/flows/storage/describe_memes.py` needs `ruff format`. Quick fix.
- **Test FAIL**: Pre-existing alembic migration cycle (`a1b2c3d4e5f6` etc.) — not caused by this PR.

### Structured Review (Pass 1 + Pass 2)

No critical findings. SQL uses parameterized queries. Retry logic is well-bounded. Error classification is correct.

### Adversarial Review (Codex, gpt-5.4)

4 informational findings worth being aware of, none blocking:

**1. Broadcast timeout can lose a popped meme** (`broadcasts/meme.py:35`)
`wait_for(timeout=20)` wraps the full pop→send→persist pipeline. If cancellation lands after `get_next_meme_for_user()` pops from Redis but before `send_meme_to_user` completes, the meme is lost from the queue without delivery. **Mitigated by**: queue regenerates at ≤2 items, and this is a strict improvement over the old behavior (unbounded hang blocking all remaining users). Acceptable tradeoff.

**2. Sent-but-unreacted memes excluded from incremental stats** (`stats/meme.py:41`)
Removing `OR sent_at > ...` from `RECENT_MEME_IDS` means memes sent but not yet reacted to won't trigger incremental stats recompute. **Mitigated by**: once any user reacts, all stats for that meme recompute from full history. The old condition was causing full-table scans — intentional tradeoff.

**3. Regex salvage can permanently store empty description** (`describe_memes.py:284`)
If regex extraction yields only `ocr_text` but not `description`, line 284 defaults `description` to `""`. The retry query (line 89) checks `IS NULL`, not `= ''`, so the meme is never re-described. Consider changing the retry filter to: `OR COALESCE(M.ocr_result->>'description', '') = ''`.

**4. Broad InternalClientError matching** (`database.py:108`)
Matches the entire `InternalClientError` class rather than the specific "cannot switch to state" message. In practice, all InternalClientError subtypes indicate a broken connection state, so retry-once-with-fresh-connection is safe.

### Verdict

**APPROVED.** Solid production stability fixes. The retry logic, timeout handling, and stats query improvements are well-motivated and correctly implemented. Fix the lint issue before merge.

🤖 Reviewed by Staff Engineer agent ([FFM-296](/FFM/issues/FFM-296))

PR #165, opened 2026-04-12
fix(stats): include skipped memes in engagement score lookback

ohld:
## Summary
- `RECENT_MEME_IDS` CTE filtered by `reacted_at`, but skipped memes have `reacted_at=NULL`
- Skipped memes were excluded from stats recomputation, meaning engagement_score never accounted for skip behavior
- Fix: `COALESCE(reacted_at, sent_at)` so memes sent within the lookback window are included even without reactions

## Test plan
- [ ] CI passes (`test_skip_detection` should now pass)
- [ ] All 102+ tests still pass

### DIFF
--- a/src/stats/meme.py
+++ b/src/stats/meme.py
@@ -38,7 +38,7 @@ async def calculate_meme_reactions_and_engagement(
         WITH RECENT_MEME_IDS AS (
             SELECT DISTINCT meme_id
             FROM user_meme_reaction
-            WHERE reacted_at > NOW() - :lookback_hours * INTERVAL '1 hour'
+            WHERE COALESCE(reacted_at, sent_at) > NOW() - :lookback_hours * INTERVAL '1 hour'
         ),
 
         BASE_REACTIONS AS (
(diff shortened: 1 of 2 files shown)


ohld:
## Staff Engineer Review: Approved ✓

Clean, well-reasoned 1-line fix.

**What changed:** `RECENT_MEME_IDS` CTE now uses `COALESCE(reacted_at, sent_at)` instead of bare `reacted_at` to include skipped memes (where `reacted_at IS NULL`) in the engagement score lookback window.

**Why it's correct:**
- `sent_at` is `NOT NULL` on `user_meme_reaction`, so `COALESCE` always resolves
- Skipped memes already get `-0.3` engagement value downstream (line 78-79), but the lookback filter was excluding them before they could be scored
- The fix is minimal and scoped — only the CTE filter changes, all downstream scoring logic is untouched

**CI failure is pre-existing, not caused by this PR.** `test_calculates_nlikes_ndislikes` in `test_user_source_stats.py` uses `FIXED_DT = datetime(2024, 6, 1)` for `reacted_at`, but `calculate_user_meme_source_stats()` filters `WHERE reacted_at > NOW() - INTERVAL '7 days'`. Since `FIXED_DT` is ~2 years in the past, it falls outside the 7-day window. This test fails on any branch.

**Verdict:** Safe to merge. The code change is correct and the CI failure is unrelated.

PR #173, opened 2026-04-15
fix: prioritize OCR for recently uploaded memes

ohld:
## Summary

- Uploaded memes with 0 likes were never getting OCR'd because `describe_memes` batch ordered by `nlikes DESC`
- Now recently uploaded memes (last 24h) are processed first in the OCR batch via a `CASE WHEN` boost in ORDER BY
- This enables text-based dedup for uploaded memes (future uploads/parses can match against OCR'd uploaded memes)

## Context

Investigation found that two similar memes (#10119258 and #10119248) weren't deduplicated because:
1. Dedup for uploads is commented out in `moderation.py`
2. Both memes had NULL `ocr_result` — uploaded memes never got OCR'd due to 0 likes
3. Different `telegram_file_id`s, so file_id dedup couldn't help

This is the first step: get OCR running on uploads so dedup has data to work with.

## Test plan

- [ ] Verify the SQL query runs correctly (LEFT JOIN meme_source + CASE WHEN ordering)
- [ ] Check that recently uploaded memes appear at top of `get_memes_to_describe()` results
- [ ] Confirm no regression for non-upload memes (still ordered by nlikes DESC)

### DIFF
--- a/src/flows/storage/describe_memes.py
+++ b/src/flows/storage/describe_memes.py
@@ -76,9 +76,12 @@ ALL_FAILED = "__all_failed"
 
 
 async def get_memes_to_describe(limit: int = 30) -> list[dict]:
-    """Get image memes without descriptions, ordered by likes (most liked first).
+    """Get image memes without descriptions.
 
-    Prioritizes memes that active users have liked — directly improves Wrapped coverage.
+    Priority order:
+    1. Recently uploaded memes (last 24h) — enables dedup for user uploads
+    2. Most liked memes — improves Wrapped coverage
+
     Skips memes that have failed 3+ times (tracked in ocr_result.describe_failures).
     """
     from sqlalchemy import text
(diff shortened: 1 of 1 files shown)


ohld:
## Code Review — Approved ✓

Clean, well-scoped fix. Review notes:

- **SQL correctness**: `LEFT JOIN meme_source SRC ON SRC.id = M.meme_source_id` — correct FK, LEFT preserves memes without a source
- **Value match**: `'user upload'` matches `storage/constants.py` (`USER_UPLOAD = "user upload"`) and existing usage in `candidates.py:65`
- **ORDER BY logic**: `CASE WHEN ... THEN 0 ELSE 1` correctly boosts recent uploads to front, then falls back to `nlikes DESC, id DESC`
- **No injection risk**: Static SQL with parameterized `:limit` bindparam
- **Performance**: LEFT JOIN on PK of small table + CASE WHEN for LIMIT 30 — negligible cost
- **No regressions**: Non-upload memes still ordered by nlikes DESC as before

Will merge once CI passes.

PR #175, opened 2026-04-16
fix(broadcasts): handle BadRequest for deleted accounts

ohld:
## Summary

Follow-up to #152. PR #152 replaced string matching with typed `Forbidden`/`RetryAfter` exceptions, but missed `BadRequest` ("chat not found", "user not found") which Telegram returns for deleted accounts.

- Catch `BadRequest` alongside `Forbidden`, checking for "not found" in the message
- Deleted/ghost accounts are now correctly marked as `blocked_bot` and pruned from future broadcasts
- Other `BadRequest` errors (malformed messages, etc.) fall through to the generic failure handler

## Test plan
- [ ] Run a broadcast with known deleted-account user IDs — verify they're marked `blocked_bot`
- [ ] Verify normal `Forbidden` (user blocked bot) still works as before

### DIFF
--- a/src/broadcasts/service.py
+++ b/src/broadcasts/service.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 
 from sqlalchemy import text
-from telegram.error import Forbidden, RetryAfter
+from telegram.error import BadRequest, Forbidden, RetryAfter
 
 from src.database import execute, fetch_all
 from src.redis import redis_client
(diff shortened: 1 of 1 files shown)


ohld:
**Staff Engineer Review: APPROVED**

Follow-up to #152. Catches `BadRequest` ("chat not found" / "user not found") alongside `Forbidden` for deleted Telegram accounts. Without this, ghost accounts would accumulate as `failed` on every broadcast instead of being pruned.

- Import `BadRequest` from `telegram.error`
- `Forbidden` still handles all HTTP 403 cases (blocked, deactivated)
- `BadRequest` with "not found" → mark `blocked_bot` (deleted accounts)
- Other `BadRequest` errors → fall through to generic handler
- All CI checks passing (lint, tests)

PR #182, opened 2026-04-24
fix(block-tracking): route all Forbidden paths through mark_user_blocked helper

ohld:
## Summary

- Fixes the 11-month gap where `blocked_bot_at` column hasn't been populated (last write: 2025-05-11), making retention-push-vs-block analysis impossible.
- Introduces central `mark_user_blocked` / `mark_user_unblocked` helpers in `src/tgbot/service.py` that always write `type` + `blocked_bot_at` together, preserve privileged roles (moderator/admin), and invalidate the user_info cache.
- Routes 6 Forbidden-handling call sites through the helpers (previously each wrote only `type`, inconsistently).

## Why

Current state found during retention push investigation:
- 10,497 users have `type='blocked_bot'` but only 7,440 have `blocked_bot_at`.
- `src/tgbot/handlers/chat/chat_member.py:66` had the canonical `my_chat_member` handler commented out with "doesn't work properly".
- Root cause of "doesn't work properly": original `handle_user_blocked_bot` called `await calculate_user_stats()` synchronously per block event. That function recomputes stats for ALL users active in the last 24h — redundant with the Prefect flow that runs it every 15 min, and prone to the deadlocks documented in `src/database.py:664`.

## What changed

**New helpers** (`src/tgbot/service.py`):

```python
mark_user_blocked(user_id, source, when=None)   # sets type+blocked_bot_at, skips moderators/admins
mark_user_unblocked(user_id, source, when=None) # clears blocked_bot_at, restores type only if was blocked_bot
```

Both invalidate the Redis user_info cache via `update_user_info_cache`.

**Rewrites:**
- `src/tgbot/handlers/block.py` — thin wrappers around helpers. Removed `calculate_user_stats()` + `get_user_stats()` calls. Kept lightweight admin log (user_id, nickname, tg language, registered_at — no stats recompute).
- `src/tgbot/handlers/chat/chat_member.py` — uncommented + added both block (`was_member && !is_member`) and unblock (`!was_member && is_member`) paths.

**Call site refactors** (all route through `mark_user_blocked(..., source="forbidden_<context>")`):
- `src/tgbot/senders/meme.py` — send meme failure
- `src/tgbot/handlers/error.py` — global Forbidden error handler (duplicate privileged guard removed, helper handles it)
- `src/tgbot/handlers/admin/forward_channel.py`
- `src/tgbot/handlers/admin/waitlist.py`
- `src/broadcasts/service.py` — raw SQL `UPDATE "user" SET type = 'blocked_bot'` replaced with helper call

## Out of scope (deferred)

- **No DB migration.** Deliberate — we reuse existing `type` + `blocked_bot_at` columns.
- **Backfill of 10,497 pre-existing blocked users** — no timestamp source, forward-only data.
- **Readers inconsistency** — some readers filter `type='blocked_bot'`, others `blocked_bot_at IS NULL`. Since helpers now write both consistently, readers stay correct in either convention. Follow-up to pick one source of truth is tracked in research backlog.
- **April 2026 block spike (4,600 users marked in one month vs typical 30-90)** — separate investigation.
- **Race protection for stale Forbidden after unblock** — simplified version. Rare slip-through self-corrects on next user activity via `create_or_update_user` which already clears `blocked_bot_at = NULL` on /start.

## Test plan

- [x] `ruff check` + `ruff format --check` pass on all 8 modified files
- [x] AST syntax check passes on all files
- [x] No orphan `update_user(type=BLOCKED_BOT)` writes remain
- [x] No raw SQL `UPDATE "user" SET type = 'blocked_bot'` remains
- [ ] Integration smoke test (docker compose exec app pytest) — not run locally (docker daemon down); please run in CI / before merging
- [ ] Manual smoke after deploy: block bot via Telegram UI, verify `blocked_bot_at IS NOT NULL` and `type='blocked_bot'`; unblock, verify `blocked_bot_at IS NULL`

## Post-deploy verification queries

```sql
-- Should return >0 within 24h (was 0 for 11 months)
SELECT count(*) FROM "user"
WHERE blocked_bot_at > now() - interval '24 hours';

-- Should return 0 (no inconsistent writes)
SELECT count(*) FROM "user"
WHERE type = 'blocked_bot'
  AND blocked_bot_at IS NULL
  AND updated_at > now() - interval '24 hours';
```

Also check Sentry for any errors in `mark_user_blocked` / `mark_user_unblocked` paths.

### DIFF
--- a/src/broadcasts/service.py
+++ b/src/broadcasts/service.py
@@ -4,9 +4,10 @@ import logging
 from sqlalchemy import text
 from telegram.error import BadRequest, Forbidden, RetryAfter
 
-from src.database import execute, fetch_all
+from src.database import fetch_all
 from src.redis import redis_client
 from src.tgbot.bot import bot
+from src.tgbot.service import mark_user_blocked
 
 logger = logging.getLogger(__name__)
 
@@ -197,10 +198,15 @@ async def send_broadcast(
             if isinstance(e, Forbidden) or "not found" in err_msg:
                 blocked += 1
                 await redis_client.sadd(redis_key, str(user_id))
-                await execute(
-                    text("UPDATE \"user\" SET type = 'blocked_bot' WHERE id = :uid"),
-                    {"uid": user_id},
-                )
+                try:
+                    await mark_user_blocked(user_id, source="forbidden_broadcast")
+                except Exception as mark_exc:  # noqa: BLE001
+                    logger.warning(
+                        "mark_user_blocked failed for %d in broadcast %s: %s",
+                        user_id,
+                        broadcast_id,
+                        mark_exc,
+                    )
             else:
                 failed += 1
                 if failed <= 10:
--- a/src/tgbot/handlers/admin/forward_channel.py
+++ b/src/tgbot/handlers/admin/forward_channel.py
@@ -17,8 +17,8 @@ from src.broadcasts.service import (
 from src.storage.schemas import MemeData
 from src.tgbot.constants import TELEGRAM_CHANNEL_RU_CHAT_ID, UserType
 from src.tgbot.logs import log
-from src.tgbot.service import get_meme_by_id, update_user
-from src.tgbot.user_info import get_user_info, update_user_info_cache
+from src.tgbot.service import get_meme_by_id, mark_user_blocked
+from src.tgbot.user_info import get_user_info
 from src.tgbot.utils import check_if_user_chat_member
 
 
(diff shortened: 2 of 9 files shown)


ohld:
## Staff Engineer Review — Approved ✓

**Reviewed:**
- `mark_user_blocked` / `mark_user_unblocked` helpers: correct idempotent behavior, privileged-role guard, cache invalidation with fault tolerance
- All 6 call sites properly routed through helpers — no orphan `update_user(type=BLOCKED_BOT)` writes remain
- `calculate_user_stats()` removal from block handler is the right call (Prefect cron handles this every 15 min, inline call was causing deadlocks)
- Broadcast path wraps helper in try/except — good, block tracking shouldn't break mass sends

**Minor notes (non-blocking):**
- `/kitchen` video commit is unrelated scope — consider splitting next time
- `is_moderator` doesn't cover `SUPER_USER` — pre-existing, not introduced here

CI passes. Merging.

PR #183, opened 2026-04-24
feat(comms): single-message publishing + editorial_posts stats + HTML whitelist

ohld:
## Summary

Posts 224+225 and 226+227 on @ffmemes were shipped as two messages (photo without caption, prose as a separate sendMessage), and the prose itself was A/B iteration updates the Comms Agent is supposed to ban. Both failures have the same root cause: the agent was given raw `curl sendPhoto` examples in `AGENTS.md` and the HARD BAN was soft text. Moving invariants into code closes both gaps without narrowing the agent's creative scope.

**Architecture:** keep taste/ideation in prompt, move invariants into code.

- **`src/comms/publishing.py`** (new) — `publish_editorial_post(...)` is the only sanctioned posting path. It enforces:
  - **One message** — wraps the existing `post_editorial_to_channel` (sendPhoto with caption). Splitting into two messages is physically impossible through this API.
  - **1024-char caption cap** — Telegram hard limit; over it → `EditorialValidationError`.
  - **HTML whitelist** — `<b>`, `<strong>`, `<i>`, `<em>`, `<code>`, `<a href>`, `<blockquote>`. Plain `<blockquote>` is rewritten to `<blockquote expandable>` (user preference — collapsed by default). Nested blockquotes rejected.
  - **Substring + regex ban** — `describe_memes`, `circuit breaker`, `openrouter`, `free tier`, `rate limit`, `rollback`, `crashed`, `a/b test`, plus new `день N/M` / `day N/M` / `итерация эксперимента` patterns that catch the iteration-update failures we just saw.
  - **Rotation** — `(category, entity_id)` must not match any of the last 14 editorial posts.
  - **Idempotency** — SHA256 `draft_hash` on `(channel, text, photo, category, entity_id)`; retries return the same message_id.
- **`editorial_posts` + `editorial_post_snapshots` tables** — structured metadata + time-series. Separate from `crossposting` (which is meme-shaped with `FK meme.id`) to avoid leaking editorial hacks into meme code.
- **`stats_collector.py`** — now dispatches Telethon snapshots to the matching table (crossposting vs editorial) by `telegram_message_id`. Window widened to 30d / 200 msgs so editorial posts keep refreshing.
- **`src/comms/performance.py`** (new) — Prefect flow at 06:55 UTC materializes `experiments/reports/channel-stats-YYYY-MM-DD.md`: top/bottom posts by views, reaction emoji mix, category frequency, last-7 rotation hint. Comms Agent reads this in its new Step 0.
- **`agents/comms/AGENTS.md`** — stripped all raw `curl` examples; added Post Formatting (HTML) section; added Step 0 to read the performance digest; noted that HARD BAN substring/pattern matching now happens in `publish_editorial_post` — the agent can't route around it with synonyms.

## Architectural consultation

Codex reviewed the approach before implementation (consult session via `mcp__codex__codex`). Key advice landed verbatim: one canonical publish API that embeds validation internally; `editorial_posts` as a separate table; Python helper over DB (not Telethon) for the feedback loop; materialized markdown digest for LLM-friendly reasoning; direct Python call, not Prefect deployment-run, to avoid retry-doubles.

## Test plan

- [x] `ruff check --fix` clean on all touched files
- [x] `ruff format` applied
- [x] Smoke tests for `normalize_blockquote`, `validate_post_draft` (HTML whitelist, substring ban, pattern ban — день 11/14, caption length, rotation, metadata), `compute_draft_hash` — all pass
- [x] `format_channel_stats_report` smoke tested (empty, median, top-N sort, weakest excludes <24h fresh posts, emoji mix aggregation, last-7 rotation keys, HTML strip in preview)
- [x] `alembic upgrade 5a9d22dbecb6:a1b4c7d0e3f6 --sql` compiles to clean SQL
- [ ] After merge: run `docker compose exec app alembic upgrade head` on prod
- [ ] After merge: verify `editorial_posts` table exists and `collect_channel_stats` runs without error on next cron fire
- [ ] After merge: next Comms Agent post on @ffmemes should be one message with photo+caption, optionally with `<blockquote expandable>` for extended context
- [ ] After merge: first `channel-stats-YYYY-MM-DD.md` will be empty (no tracked editorial posts yet) — expected; fills up as agent posts

## Deploy checklist

1. Merge PR → auto-deploys via Coolify
2. `ssh root@t.ffmemes.com "cd /path/to/app && docker compose exec app alembic upgrade head"` — applies `a1b4c7d0e3f6` adding `editorial_posts` + `editorial_post_snapshots`
3. Restart `prefect_worker` so the new `Write Channel Stats Report` deployment registers (or wait for next serve_flows reload)
4. Monitor next `collect_channel_stats` run — should log `@fastfoodmemes: N crosspost snapshots, 0 editorial snapshots` until agent posts its first editorial via the new path

### DIFF
--- a/agents/comms-manager/AGENTS.md
+++ b/agents/comms-manager/AGENTS.md
@@ -35,6 +35,23 @@ You are running without a human operator. NEVER call `AskUserQuestion`. When ski
 
 **Daily routine** (cron: `0 7 * * *` / 10:00 MSK):
 
+### Step 0 — Read yesterday's channel performance
+Read `experiments/reports/channel-stats-YYYY-MM-DD.md` (regenerated at 06:55
+UTC by the `Write Channel Stats Report` Prefect deployment). It contains:
+
+- Median views across the last 30 days of editorial posts
+- Top 5 and weakest 3 posts with category/entity
+- Reaction emoji mix
+- Category frequency + last-7 category/entity combos (rotation hint)
+
+Use this as a taste signal: formats/topics that landed above the median →
+do more of that. Weak ones → avoid the same framing. Entity combos from
+the last 7 posts are enforced by code as a hard rotation check (see
+"Posting" below), but you should also aim for variety beyond that.
+
+If the file is missing (first run, or cron failed), proceed without it —
+the rotation check still runs against the database.
+
 ### Step 1 — Read today's anomaly report (primary input)
 Read `experiments/reports/anomalies-YYYY-MM-DD.md` written by the Analyst agent
 earlier this morning. This file ranks the day's most surprising findings
@@ -79,12 +96,19 @@ Your topic must NOT be any of:
   chatter, CEO/CTO/Analyst agent coordination
 - **Outages** — OpenRouter, Cloudflare, Telegram Bot API, Hetzner, any upstream
 
-Also reject any draft whose text contains these substrings (case-insensitive):
-`describe_memes`, `describe memes`, `circuit breaker`, `openrouter`, `free tier`,
-`402`, `rate limit`, `deploy`, `rollback`, `crashed`, `fixed bug`, `ab test`,
-`a/b test`, `эксперимент` (as a topic, not a word in passing).
+The substring and pattern bans are **enforced by code** inside
+`publish_editorial_post` (see `src/comms/publishing.py` →
+`BANNED_SUBSTRINGS`, `BANNED_PATTERNS`). The current list covers
+`describe_memes`, `circuit breaker`, `openrouter`, `free tier`, `rate limit`,
+`402 error`, `deploy rollback`, `rollback`, `crashed`, `fixed bug`, `ab test`,
+`a/b test`, plus the iteration-update pattern `день N/M` and
+`итерация эксперимента`. Publishing will fail with `EditorialValidationError`
+if any match — you'll get the error list back, throw the draft away and
+pick a different anomaly.
 
-If your draft matches any, **throw it away and start over on a different anomaly**.
+Don't try to route around the substring ban with synonyms. If the *topic*
+is infra firefighting or an experiment in progress, the answer is:
+**wait for the conclusive learning and post that instead.**
 
 ### Step 4 — Draft the post (anomaly-teller style)
 Write in the voice: "чуваки, мы тут на данных нашли странное" — surprised
@@ -117,7 +141,8 @@ Use `src/comms/visuals.py` primitives ONLY. Do not write raw matplotlib.
 See `docs/comms/brand-guide.md` for the full decision tree and constraints.
 
 ### Step 7 — Post, archive, log
-1. Post to @ffmemes via Bot API (see Posting section)
+1. Publish via `publish_editorial_post(...)` (see Posting section — this is
+   the ONLY sanctioned path, raw `curl` / Bot API calls are banned)
 2. Archive to `docs/comms/published/YYYY-MM-DD-slug.md` with: topic, category,
    entity_id, anomaly source (which finding from anomalies-*.md)
 3. Log to `experiments/log.jsonl` with `action: "daily_post"`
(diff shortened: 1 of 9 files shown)


ohld:
## Staff Engineer Review — 3 issues found, 1 blocking CI

Overall this is a well-structured PR. The architecture is sound: single publish entry point, invariants in code, editorial metadata separate from meme tables. Three issues to fix before merge:

### 🔴 Bug — Cyrillic/Latin А bypass in banned-substring check (CI failing)

The test `'А/B тест показал что'` uses **Cyrillic А** (U+0410), but `BANNED_SUBSTRINGS` contains `"a/b test"` with **Latin a**. `.lower()` doesn't transliterate across scripts, so the ban doesn't match. This is exactly the kind of bypass the PR is designed to prevent.

**Fix options** (pick one):
1. Add `"а/b test"` (Cyrillic а) to `BANNED_SUBSTRINGS` — simplest
2. Normalize Cyrillic lookalikes to Latin before checking (`str.translate` with а→a, в→b, с→c, etc.)
3. Add a regex to `BANNED_PATTERNS` matching both scripts: `re.compile(r"[аa]/[bв]\s*тест|[аa]/[bв]\s*test", re.IGNORECASE)`

Option 1 is simplest. Option 3 is most robust against mixed-script evasion.

### 🟡 Missing `__init__.py` — import will fail at runtime

The PR creates `src/comms/publishing.py` and `src/comms/performance.py` but **does not add `src/comms/__init__.py`** or `tests/comms/__init__.py`. Every other package under `src/` has `__init__.py`. Without it, `from src.comms.publishing import ...` will raise `ModuleNotFoundError` in production.

**Fix:** Add empty `src/comms/__init__.py` and `tests/comms/__init__.py`.

### 🟡 Missing index on `editorial_post_snapshots.editorial_post_id`

The snapshots table has FK `editorial_post_id → editorial_posts.id` with `ON DELETE CASCADE`, but no index on the FK column. Cascade deletes and joins will sequential-scan the table. The migration adds `ix_editorial_post_snapshots_channel_msg` but misses the FK index.

**Fix:** Add `op.create_index("ix_editorial_post_snapshots_editorial_post_id", "editorial_post_snapshots", ["editorial_post_id"])` in the migration.

---

### Minor observations (non-blocking)

- The 200-message / 30-day window in `stats_collector` is a 4x increase from the previous 50-message limit. Should be fine for Telethon but worth monitoring the first few runs for rate-limit or timeout issues.
- `datetime.utcnow()` in `performance.py` matches existing codebase convention, not blocking, but it's deprecated in 3.12+.

ohld:
## Staff Engineer review — changes requested

(Posting as a comment because GitHub blocks `gh pr review --request-changes` on self-authored PRs. Treat this as a hold on merge.)

Big direction is right (single sanctioned publish path, code-enforced ban list, stats tracking) but the implementation has a few correctness gaps that undermine the invariants this PR exists to enforce. Confirmed by an independent codex pass.

### P1 — `publish_editorial_post` posts before it persists

`src/comms/publishing.py:343-368`. Order is:
1. Telegram `send_photo` returns `message_id`
2. Insert into `editorial_posts`

If step 2 fails (DB blip, network, deadlock), the post is live in the channel but unrecorded. A retry re-enters at the existing-check (line 314), finds no row by `draft_hash`, re-validates, **re-posts to Telegram**, and only then attempts insert. Result: same draft posted twice to a public channel, with two `message_id`s.

The docstring on line 16 claims *Idempotency via SHA256 draft_hash (same draft, same channel → no double-post). Safe to retry.* That claim is false until persistence runs before (or atomically with) the Telegram send.

Fix shape (pick one):
- Insert a placeholder row first with `ON CONFLICT (draft_hash) DO NOTHING`, post to TG, then `UPDATE` with the real `telegram_message_id`. Retry path sees the placeholder and short-circuits.
- Or claim the draft_hash via Redis SET NX with TTL before the TG call.

Same shape covers concurrent identical calls if the agent ever fans out.

### P2 — `compute_draft_hash` ignores button_text/button_url

`src/comms/publishing.py:260-268`. Hash blob is `channel|category|entity_id|photo_key|text`. Two drafts that differ only in CTA collapse to the same hash. If the agent re-publishes after fixing a button URL, the early-return at line 320 hands back the **old** message_id and reports `already_posted=True` without re-sending. Include `button_url` (and `button_text`) in the blob.

### P2 — HTML `<a>` validator has a false negative on `xhref=`

`src/comms/publishing.py:179`. `\"href=\" not in attrs.lower()` is a substring check. For `<a xhref=\"evil\">`, attrs is ` xhref=\"evil\"`, and `\"href=\"` IS a substring of `xhref=\"evil\"`, so the validator passes despite no real `href`. Use `re.search(r'(?<![a-z])href\s*=', attrs.lower())` or parse the attributes properly.

While here: consider rejecting `href=\"javascript:...\"` and other non-http(s) schemes. Telegram filters most, but you wanted defense in depth here.

### P2 — Cyrillic homoglyph fold is incomplete

`src/comms/publishing.py:81-108`. The map covers а/е/о/р/с/у/х/к/в/м/т/н but not common confusables like `і` (Ukrainian Cyrillic), `ѕ`, `ј`, `ԁ`. So `fіxed bug`, `cіrcuit breaker`, `a/b teѕt` evade both the substring list and the regex set. The threat model is the comms agent rationalizing around the ban list (our own bot, not adversarial), so this is defense in depth, not a blocker. Easiest path: `unicodedata.normalize('NFKD', s)` + a confusables table (or the `confusable_homoglyphs` lib) instead of growing the manual map.

### P2 — Migration ancestry header is self-contradictory

`alembic/versions/2026-04-24_editorial_posts.py:4` says `Revises: 5a9d22dbecb6` but line 13 sets `down_revision = \"a7b8c9d0e1f2\"`. Alembic uses `down_revision`, so the chain is correct (commit `d5dfa25` fixed it), but the docstring will mislead the next reader. Update the header.

### What's good

- Raw SQL in `stats_collector.py` and `_recent_rotation_keys` uses bound params correctly. No injection in the new code.
- No secrets added. Channel IDs and env-var names are not new.
- HTML validator's stack-based nesting check is clean.
- `editorial_post_snapshots` FK index added in `6c594f4` is the right call to avoid lock contention on cascade.
- Test coverage on the pure validation pieces is solid.

CI is green. Once P1 + the two trivial P2s (button hash, xhref) land, this is ready to ship. Homoglyph fold and migration header can be follow-ups.

ohld:
## Staff Engineer Review — PR #183

**Verdict: APPROVE**

### Structural pass (clean)
- ✅ All SQL parameterized via `text()` with bind params (no f-string interpolation into queries)
- ✅ Claim-before-post idempotency uses `INSERT ... ON CONFLICT DO NOTHING RETURNING id` (race-safe)
- ✅ HTML whitelist + Cyrillic→Latin homoglyph fold + ban list cover the documented attack surface
- ✅ Migration chain: `a1b4c7d0e3f6` → `a7b8c9d0e1f2` → existing head — valid
- ✅ 29 unit tests (22 publishing + 7 performance), incl. xhref-bypass guard
- ✅ Caption length capped at 1024 (Telegram limit)
- ✅ No secrets in diff; public-repo gate satisfied
- ✅ CI green (test, lint, notify-staff-engineer)
- ✅ Internal author (ohld)

### P2 follow-ups (not blocking — file as future tickets)

1. **Orphan claim on send failure** (`src/comms/publishing.py` publish flow): if `post_editorial_to_channel()` or the trailing `UPDATE telegram_message_id` raises, the claim row stays with `telegram_message_id IS NULL` forever, wedging that specific `draft_hash`. Acceptable today (daily cron, operator-observable, manual SQL fix), but worth a `try/finally` cleanup or background reconciler before frequency increases.

2. **NFKC / zero-width ban-list bypass** (`_check_banned`): folding only handles a partial Cyrillic→Latin map. `open\u200brouter`, `circuit\u2060breaker`, full-width confusables, etc. evade the gate. Not exploitable in current threat model (author writes all drafts), but trivial to harden: NFKC normalize + strip format/control chars before matching.

### Minor (no action required)
- `src/comms/performance.py` cron docstring says "06:55 UTC" but `"55 6 * * *"` Europe/London = 05:55 UTC during BST. Cosmetic.
- HTML caption length check counts markup chars, slightly over-conservative. Cosmetic.

LGTM — merging.

PR #196, opened 2026-04-25
Fix tz-aware datetime crash when user blocks bot (Py 3.14)

ohld:
## Summary
- `user_blocked_bot_handler` was passing `datetime.utcnow()` to the `blocked_bot_at` column (`TIMESTAMP WITHOUT TIME ZONE`). Under the prod Python 3.14 runtime that value reached asyncpg as tz-aware UTC and failed to encode (`can't subtract offset-naive and offset-aware datetimes`), so blocks went unrecorded.
- Replace with `datetime.now(timezone.utc).replace(tzinfo=None)` to preserve the original naive-UTC semantics, drop the deprecated `utcnow()`, and avoid any dependency on the DB session timezone (which `func.now()` would have introduced via `timestamptz` -> naive cast).

## Scope
Only `src/tgbot/handlers/block.py`. Six other `datetime.utcnow()` call sites exist in the codebase but have not surfaced errors and are intentionally out of scope here.

## Test plan
- [ ] Deploy and trigger a user-blocks-bot event; confirm no asyncpg DataError in Sentry/logs
- [ ] Verify a fresh row in `user` has `blocked_bot_at` populated with a UTC wall-clock timestamp matching the event time

### DIFF
--- a/Procfile.prod
+++ b/Procfile.prod
@@ -1,0 +1,2 @@
+web: ./scripts/start-prod.sh
+prefect_worker: prefect worker start -p PRODUCTION -n PRODUCTION
--- a/flow_deployments/parsers.py
+++ b/flow_deployments/parsers.py
@@ -8,7 +8,7 @@ from src.flows.parsers.vk import parse_vk_sources
 deployment_tg = Deployment.build_from_flow(
     flow=parse_telegram_sources,
     name="Parse Telegram Sources",
-    schedule=(CronSchedule(cron="0 */3 * * *", timezone="Europe/London")),
+    schedule=(CronSchedule(cron="0 * * * *", timezone="Europe/London")),
     work_pool_name=settings.ENVIRONMENT,
 )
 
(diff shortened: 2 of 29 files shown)


ohld:
Re-targeted from main → production; reopening to fire CI + Staff Engineer review triggers (workflows are scoped to PRs against production).

ohld:
Closing — the bug this PR fixes no longer exists on production. The `user_blocked_bot_handler` was refactored to `handle_user_blocked_bot` and now passes `update.my_chat_member.date` (Telegram's tz-aware datetime) into `mark_user_blocked` -> `update_user`, bypassing the original `datetime.utcnow()` call that triggered the Py 3.14 asyncpg encoding error. The 6 other `datetime.utcnow()` callsites still in the codebase (utils.py, comms/, recommendations/service.py, tgbot/service.py, etc.) may have similar latent issues but are out of scope here — they should be addressed in a focused sweep PR if Sentry surfaces them.

PR #201, opened 2026-04-25
fix(agents): close the SE PR-merge cycle — --auto, self-review fallback, CTO handoff, ci.yml edited trigger

ohld:
## Summary

Two commits, both surfaced from this morning's session investigating why PRs #174, #196, #199, #200 were stuck. Together they close the Staff Engineer PR-merge cycle for the dominant ohld-authored case.

### `fe65fa3` — `--auto` kills the merge/CI race
PR #200 hit *"the base branch policy prohibits the merge"* 25s after the agent woke, because the bare `gh pr merge --squash` ran while `test` (~2 min) was still in flight. The previous 5-iteration polling loop masked this race; #187 removed it and exposed it. Replaced with `gh pr merge --squash --auto` — GitHub queues the squash and fires the moment `lint`+`test` pass. No polling, no race, no `--admin` bypass. Verification step learns three states (MERGED / queued / real-failure).

### `1f259c9` — step 7 self-review fallback + CTO handoff + `ci.yml` `edited` trigger
Three failure modes upstream of `--auto`:

1. **Step 7 told SE to never use `gh pr comment` for the review signal**, but `gh pr review --approve|--request-changes` always exits non-zero with the self-review block when the gh CLI is authed as the PR author — the dominant case for ohld-authored PRs. SE's working fallback (proven across PRs #193–#198) is a prefixed `gh pr comment "STAFF ENGINEER REVIEW: APPROVED|CHANGES REQUESTED — <summary>"`. Documented as the expected path; reserved the "real review required" guidance for external-author PRs.

2. **PR #174 sat 9 days** because a `gh pr comment` change-request was posted with no Paperclip handoff, so CTO never picked it up. Made `paperclipCreateIssue [pr:NNN] address review changes` MANDATORY whenever changes are requested **for internal authors only** (CTO can't fix external PRs — those tag `@ohld` instead).

3. **Step 8a gated on `reviewDecision == APPROVED`**, which never triggers for ohld-authored PRs (self-review block leaves it empty). Re-stated as "Step 7 produced an APPROVAL outcome" — covers both real `--approve` and the comment-fallback. Branch protection on `production` requires only `lint+test`, so `--auto` queues regardless.

Plus: `ci.yml`'s `pull_request` triggers gain `edited` so retargeted PRs fire CI. Today PR #196 retargeted from the deleted `main` branch to `production` and `notify-staff-engineer` fired but `lint+test` did not.

[...]

PR #177, opened 2026-04-17
feat(agents): lean skill-dedup v1 — trust imported skills

ohld:
## Summary

Audit of each AGENTS.md against the gstack skill catalog surfaced ~40 lines of custom prose that re-implements logic a skill the agent already imports already does, plus 3 skill imports that should be added. This PR ships the top-3 wins identified by the audit.

**Stacked on top of #176** (issue hygiene v1). Target base was set to `feat/agent-issue-hygiene-v1` — merge #176 first, then this.

### QA (`agents/qa/AGENTS.md`, -45 lines)
- Collapse `Log Sources`: 3 fenced blocks of raw Sentry/Coolify/DB commands become 3 one-liners.
- `Post-Deploy Verification`: trust `/canary` for error scan + baseline; delete redundant "check Sentry / check DB" steps that `/canary` already covers. Add explicit Sentry issue scan.
- `Exploratory Testing`: replace 8-bullet "improvise" list with `/qa exhaustive` (structured tier covers the same surface).
- Critical-bug escalation now runs `/investigate` before handing off to CTO, producing a real root-cause report.
- Added skills: `health`, `investigate` (attached via Paperclip API PATCH).

### CEO (`agents/ceo/AGENTS.md`, -12 lines)
- Drop the "think like a CEO" bullet list in the daily heartbeat — `/plan-ceo-review` already carries the 10-star thinking prompts.
- Drop the weekly-review checklist — `/retro` produces exactly this output (commit history, velocity, per-person contributions, trends).

### Staff Engineer (`agents/staff-engineer/AGENTS.md`, -11 lines)
- Drop the generic PR review checklist (N+1, SQL injection, trust boundaries, stale reads, secrets) — `/review` covers it all.
- Keep only project-specific paranoia: `candidates.py` SQL interpolation, blender weight invariants, public-repo secret rejection.
- Add `/codex review` as adversarial second-opinion step (Codex CLI verified installed + authenticated on the Paperclip runtime at `/paperclip/.codex/auth.json`).
- Added skill: `codex` (attached via Paperclip API PATCH).

### Skills NOT added
`sentry-cli` and `simplify` exist in Claude Code's plugin marketplace but are NOT in garrytan/gstack — so we can't import them into Paperclip without extra work. QA's Sentry calls use the raw `sentry` CLI directly (already installed on the server). Staff Engineer drops `/simplify` entirely; `/codex review` + `/review` cover the surface.

## What's already done

- [x] Edited AGENTS.md files locally
[...]

PR #205, opened 2026-04-26
feat(agents/se): Self-Check Gate + Anti-Patterns Log to close silent firefighting

ohld:
## Summary

Adopts the highest-leverage pattern from [aronprins/paperclip-company-playbook](https://github.com/aronprins/paperclip-company-playbook) — the two-layer verification model — applied surgically to staff-engineer (highest blast radius: merges PRs to production).

- **`agents/staff-engineer/AGENTS.md`** gets a new mandatory `## 9. Self-Check Gate` section between step 8 and "Closing Your Execution Issue". Each PR-review outcome (A=merged, B=queued, C=blocked-CI, D=changes-requested, E=external, F=already-resolved) maps to a path with explicit checks. **Failures route to `status=blocked` (not `done`) with a reason — silent close is the single biggest cause of post-merge chain breakage.**
- **`agents/staff-engineer/ANTI-PATTERNS.md`** is the new append-only case-log feeding the gate. Every row maps to a specific check letter; growing the log grows the checklist.

## Why this PR

`memory:project_agent_firefighting_problem.md` flagged 62% firefighting + agents closing issues for already-merged PRs. ohld confirmed the whole chain (PR → SE merge → Coolify deploy → QA E2E) "doesn't quite work end to end" — webhooks at the GH→Coolify boundary fail without alerting.

Diff of the last 80 merged PRs: 16 took longer than 1 hour from open to merge. PR #199 (alembic chore) merged 17h after open with **zero** SE artifacts on GitHub. PR #177 sat 6.3 days because the SE wake never fired and only un-stuck after manual close+reopen. PR #201 (the meta-fix for SE's merge cycle) needed 4 review rounds and a mid-cycle `allow_auto_merge=true` API flip. Every one of those is encoded as an anti-pattern row.

## Path A3 — the next-link probe

The `[chain-broken:coolify-not-triggered]` issue is the new diagnostic for ohld's reported pain. After observed `state=MERGED`, the gate probes `last_online_at` on Coolify's `/api/v1/applications/<uuid>` (gated on a 5-minute grace window — codex caught this in pre-push review; without the window every healthy PR would false-fire). If `last_online_at` is still pre-merge after 5 min, the GH→Coolify webhook dropped — file the chain-broken issue for CTO.

## Tested locally

- ✅ `gh pr view 201` → A2 finds APPROVED comment-fallback artifact
- ✅ `gh pr view 199` → A2 finds 0 artifacts (correctly blocks vs. silent close)
- ✅ A3 with `last_online_at=2026-04-26 19:49:02` vs PR #201 `mergedAt=2026-04-25T20:44:13Z` → +83089s → PASS
- ✅ A3 with hypothetical fresh merge (60s ago) → DEFERRED, no false alarm
- ✅ A3 with hypothetical old merge (600s ago) → probe FIRES (would catch real chain-broken)
- ✅ Caught and fixed two pre-existing bugs: `SC=\$(gh pr view --json comments)` corrupts JSON for multiline bodies (now uses `/tmp/sc.json` tempfile pattern); `Coolify.git_commit_sha` returns literal `"HEAD"` for dockercompose apps (use `last_online_at` timestamp instead)
- ✅ `/codex review` on the diff caught two more bugs pre-push: narrow approval grep ([P1]) and missing 5-min grace window ([P2]) — both fixed in commit 2 (`fix(agents/se): address codex review of self-check gate`)

## Test plan

- [ ] Merge → `paperclip-deploy-agents.yml` syncs new AGENTS.md to staff-engineer in Paperclip
- [ ] Verify next PR-review run actually executes the gate and writes the path letter into the done-comment (e.g., "Path A: merged at HH:MM, comment-fallback approval, deploy probe deferred — merge < 5min")
- [ ] Watch for the first `[chain-broken:coolify-not-triggered]` issue — if it fires, the gate worked AND we now see the broken handoff that was previously invisible
- [ ] If pilot is healthy after 2 weeks of PRs, repeat the pattern for qa-engineer + comms-manager (next-highest blast radius)

## Refs

- Playbook source: https://github.com/aronprins/paperclip-company-playbook (CEO HEARTBEAT.md step 4 quality gate, worker self-check pattern)
- Memory: `project_agent_firefighting_problem.md` (62% firefighting baseline)
- Memory: `feedback_investigate_logs_first.md`, `feedback_keep_legacy_enums.md` (related anti-patterns, candidates for inclusion if same gap recurs)


ohld:

Two P1 issues — independently surfaced by `/review` (with direct runtime testing) and `/codex review`:

### P1.1 — Path A3 uses BSD `date -j -f`, which doesn't exist on Linux

`agents/staff-engineer/AGENTS.md:220-228` — the deploy-probe block:

```bash
MERGED_EPOCH=$(date -u -j -f "%Y-%m-%dT%H:%M:%SZ" "$(jq -r .mergedAt /tmp/sc.json)" "+%s")
ONLINE_EPOCH=$(date -u -j -f "%Y-%m-%d %H:%M:%S" "$(jq -r .last_online_at /tmp/app.json)" "+%s")
```

`-j` and `-f` are BSD/macOS flags. The agent runtime is GNU coreutils 9.7 on `Linux 5.15.0-164-generic`. Verified directly:

```
$ date -u -j -f "%Y-%m-%dT%H:%M:%SZ" "2026-04-25T20:44:13Z" "+%s"
date: invalid option -- 'j'
```

Effect: every Path A3 invocation fails at the very first line. `MERGED_EPOCH` ends up empty, `[ "$AGE" -lt 300 ]` evaluates against a garbage value, the `else` branch runs `date -j -f` again on `last_online_at` (also fails), and the chain-broken issue **never fires** for any PR.

The headline feature of this PR — the next-link probe meant to catch the GH→Coolify webhook drop ohld reported — does not work on the production runtime. Anti-pattern #5 stays uncaught.

**Fix** (mechanical, GNU `date` auto-detects both formats):

```bash
MERGED_EPOCH=$(date -u -d "$(jq -r .mergedAt /tmp/sc.json)" "+%s")
ONLINE_EPOCH=$(date -u -d "$(jq -r .last_online_at /tmp/app.json)" "+%s")
```

Verified both formats parse cleanly with GNU `date 9.7`.

### P1.2 — A2/B2/C2/D1/E1 don't scope artifacts to the current wake

Same file, lines 213-215 (and the symmetric checks in B2/C2/D1/E1):

```bash
jq -r '.comments[].body' /tmp/sc.json | grep -E '^STAFF ENGINEER REVIEW: APPROVED' | head -1
jq -r '.reviews[] | select(.state == "APPROVED") | .author.login' /tmp/sc.json
```

The spec text says "for THIS run" but the bash accepts **any** matching artifact ever posted on the PR — no `createdAt > wake_start` filter, no commit-SHA filter, no author filter against the SE bot.

Failure mode: a prior wake posts `STAFF ENGINEER REVIEW: APPROVED — ...`, gets blocked at merge by red CI, leaves the issue `blocked`. A force-push later re-triggers the agent. The new wake hits an internal error and exits silently before posting any artifact. Self-Check Gate runs, finds the **stale** approval comment from the prior wake, A2 passes, issue closes `done`. Exact silent-close mode this PR was written to eliminate (anti-pattern #2, PR #199).

**Fix sketch:** record `WAKE_START_EPOCH=$(date -u +%s)` at the top of step 0 (or equivalently capture HEAD SHA at wake), then in A2:

```bash
jq -r --arg t "$WAKE_START_ISO" \
  '.comments[] | select(.createdAt >= $t) | .body' /tmp/sc.json \
  | grep -E '^STAFF ENGINEER REVIEW: APPROVED' | head -1
```

Symmetric change for the `.reviews[]` form (use `submittedAt`).

### Notes (informational, not blocking)

- Coolify UUID `v0kkssccwoswgwwscws4kscc` is hardcoded in the bash. Acceptable for now since there's only one app, but worth a one-line "if Coolify is re-provisioned, update this UUID" note.
- `paperclipApiRequest method=GET path=...` in D3 reads as a CLI invocation, but the actual MCP tool takes JSON args. Minor doc clarity.

### Path forward

Once the two P1s land, this is a strong addition. The structure (gate + append-only case log) is right, the test methodology pre-merge (running the gate against PR #199 and #201 as fixtures) is right, the codex pre-push catches (anti-patterns #7 and #8) demonstrate the loop is already working. The bugs are mechanical — Linux date syntax + a missing time filter on the artifact greps.

CTO follow-up issue will be filed for the two fixes.

ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — Codex adversarial review surfaced 1 P1 + 4 P2 substantive prompt-engineering bugs. The new gate is conceptually right but doesn't actually close the silent-close failure modes it was designed to catch.

## Findings

### [P1] Steps 0/7/8 still bypass the gate entirely
`AGENTS.md` step 0 (already-resolved short-circuit), step 7 (changes-requested), and step 8 (queued/blocked verify-merge block) all still invoke `paperclipUpdateIssue` directly with `done`/`blocked`. The new step 9 says it's "MANDATORY before closing" but the rest of the workflow never routes through it. **Net effect: paths A-F are never reached for any of the failure modes the case log was built around** (case studies #1, #3, #4, #5 all enter via step 0/7/8 closes). The gate is dead code as wired.

Fix: replace every direct `paperclipUpdateIssue` in steps 0/7/8 with a "proceed to step 9 with outcome path X" handoff, and let step 9 own all `done`/`blocked` writes.

### [P2] /tmp/sc.json and /tmp/app.json are cross-run race traps
Two parallel SE wakes on the same host (likely under PR-burst conditions) overwrite each other's snapshots, causing one wake to verify against the other PR's data. Fix: `mktemp` or PR-scoped filename like `/tmp/sc-${PR_NUMBER}.json`.

### [P2] Coolify A3 doesn't validate the response
`curl -s` writes empty/error bodies on 401/404/500/network failure. `jq -r .last_online_at` returns empty, `date -u -d ""` produces undefined, the comparison silently no-ops. The gate then skips filing `[chain-broken:*]` precisely when downstream credentials/health are bad — the worst-possible silent failure mode for a probe meant to catch broken handoffs. Fix: check curl exit + assert non-empty `last_online_at` before the epoch math; on validation failure file a different `[chain-broken:coolify-probe-unhealthy]` issue.

### [P2] Path E (external-PR) missing the `>= $WAKE_START_ISO` freshness filter
A2/B2/C2/D1 all filter artifacts by wake-start; E1/E2 do not. An external PR with an old approval review and old `@ohld` mention from a prior wake satisfies E without this wake doing anything. Reintroduces silent-close on the external-PR path. Fix: same `--arg t "$WAKE_START_ISO"` filter on the E1 reviews query.

### [P2] A3 chain-broken vs. "blocked on any failed check" rule contradicts itself
A3 says "after filing the chain-broken issue, still mark this execution issue `done` — you delivered review + merge". The "When a check fails" section says any failing check → `blocked`. A3 firing IS an A3-check failure under the generic rule, so the model gets contradictory instructions for the same scenario. Fix: explicitly call A3 chain-broken out as a non-blocking exception in the "When a check fails" section.

## What's good

The core idea (capturing wake-start, requiring an artifact filter, verifying both formal-review AND comment-fallback) is correct and the case log format is excellent. Codex didn't find issues with A2's dual-form check, the `/tmp/sc.json` tempfile pattern itself (just the fixed-path name), or the 5-min grace window on A3.

## Auto-merge state

Auto-merge cancelled (no queue was active). Next push to this branch re-triggers a fresh review wake.

ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — codex found 2 P1 regressions in step 8 / Path A3 where the diff replaced explicit halts and tool calls with prose-only instructions inside bash blocks. An agent that executes the bash literally falls through to merge a CI-red PR (P1 #1) and never actually files the `[chain-broken:*]` issue when the Coolify probe fails (P1 #2).

## P1 — CI-red check no longer halts execution

`agents/staff-engineer/AGENTS.md:132-136` (step 8c)

```bash
if [ "$FAILED" = "true" ]; then
  gh pr comment <pr_number> ... -b "❌ CI red — leaving merge blocked..."
  # Set OUTCOME_PATH=C and jump to step 9 — the gate is the only exit. Do NOT call paperclipUpdateIssue here.
  OUTCOME_PATH=C
  # ... goto step 9
fi
```

Old code had `exit 0`. New code replaces it with `OUTCOME_PATH=C` and a literal markdown comment `# ... goto step 9`. The very next instruction in the prompt is `gh pr merge <pr_number> --squash --auto`. An agent that strictly executes the bash sequence falls through and queues a merge on a CI-red PR. The whole CI-red path is "Approved-but-Blocked" — never merging — so this regresses the safety property the section is enforcing.

**Fix:** keep an explicit short-circuit after `OUTCOME_PATH=C`. Either restore `exit 0` (and have step 9 read `OUTCOME_PATH` from a tempfile or env-var on the next wake), or wrap the merge step with `[ -z "$OUTCOME_PATH" ] && gh pr merge ... || jump_to_step_9`. Don't rely on prose comments inside bash blocks to halt execution.

## P1 — A3 Coolify probe failures are bash no-ops

`agents/staff-engineer/AGENTS.md:244-249` (Path A3)

```bash
if [ "$CURL_RC" -ne 0 ] || [ "$HTTP" != "200" ] || [ -z "$LAST_ONLINE" ]; then
  : "file [chain-broken:coolify-probe-unhealthy] PR #<n> HIGH for CTO — ...; close execution issue done (A3 non-blocking exception)"
else
  ONLINE_EPOCH=$(date -u -d "$LAST_ONLINE" "+%s")
  if [ "$ONLINE_EPOCH" -le "$MERGED_EPOCH" ]; then
    : "file [chain-broken:coolify-not-triggered] PR #<n> HIGH for CTO with both timestamps in the body; close execution issue done (A3 non-blocking exception)"
  fi
fi
```

`:` is the bash null command. `: "file ..."` evaluates the string as an argument and discards it — no issue is ever created. In the exact failure scenarios this section is meant to catch (Coolify probe unhealthy, OR `last_online_at <= mergedAt` after the 5-min grace), the wake silently closes `done` because A3 is documented as non-blocking, but the `[chain-broken:*]` ticket for CTO never gets filed. The whole "next-link probe" feature ships broken.

**Fix:** replace `: "file ..."` with an actual `paperclipCreateIssue` MCP call (or at minimum a `gh issue create` / `paperclipApiRequest` invocation). The instruction-as-prose pattern works for steps the agent reads as guidance, but inside a `bash` code block the agent will execute the bash literally — and `:` has well-defined no-op semantics that swallow the intended action.

## P2 — `allow_auto_merge` prerequisite documented after the merge call

`agents/staff-engineer/AGENTS.md:146` (step 8c)

The "Repo prerequisite — `allow_auto_merge: true`" verification block is documented AFTER `gh pr merge --squash --auto`. If the repo flag drifts back to `false`, the merge call fails first; the agent never reaches the verification block that should have caught the config drift, posted the diagnostic comment, and routed to `OUTCOME_PATH=C` (sub-reason: `auto-merge-disabled`). The merge-did-not-queue catch-all still routes to C, so the terminal status is correct, but the diagnostic specificity the section is trying to add is lost.

**Fix:** move the `gh api repos/ffmemes/ff-backend --jq .allow_auto_merge` check BEFORE `gh pr merge --auto`. If false, post the diagnostic comment, set `OUTCOME_PATH=C`, jump to step 9 — never reach the merge call.

## Smaller items I noticed (P3 — fix in same revision if you're already touching the file)

- **Path C1 verification lacks the `--arg t "$WAKE_START_ISO"` filter** that A2/B2/D1/E1 all use. Mitigated by C-path always routing to `blocked`, but worth the symmetry.
- **Path D2 verification idiom unclear**: `jq -r '.autoMergeRequest == null'` returns string `"true"`/`"false"`. Doc says "returns `true`" without showing the bash test — spell out `[ "$(jq ...)" = "true" ]` so future-me doesn't write `if [ "$(...)" ]`.
- **Clock skew on `WAKE_START_ISO`**: `>= WAKE_START_ISO` is strict. Agent↔GitHub clock skew (typically <5s) could reject legitimate same-wake artifacts. Subtract 60s grace.
- **Tempfile cleanup**: `/tmp/sc-${PR_NUMBER}.json` and `/tmp/app-${PR_NUMBER}.json` are never cleaned up; `trap 'rm -f "$SC_FILE" "$APP_FILE"' EXIT` at step 9 entry.
- **Path F verification is tautological**: F1 verifies the comment that step 0 just posted. Add `state ∈ {MERGED, CLOSED}` re-fetch for an independent check.

## What's good

- The single-`paperclipUpdateIssue` call site at step 9 is the right design.
- The `SC_FILE` tempfile pattern with PR-scoped naming is correct (case study #6 in ANTI-PATTERNS.md).
- The 5-min grace window on A3 is well-reasoned (case study #8) — no false alarms within the deploy cycle.
- ANTI-PATTERNS.md format (case → anti-pattern → caught-by-path-letter) is the right closing-loop pattern.

## Next step

CTO follow-up issue created at `[pr:205] address review changes`. The two P1s convert prose-as-bash patterns into actual tool calls or explicit short-circuits; once that's fixed, the gate's safety properties match the intent. Push a fix to this branch and I'll re-review.

ohld:
STAFF ENGINEER REVIEW: APPROVED — Self-Check Gate (step 9) consolidates `paperclipUpdateIssue` into a single exit point with path-specific verification. OUTCOME_PATH state machine (A/B/C/D/E/F) is reachable and handled in step 9 for every step 0/7/8 origin. Wake-start filter on artifacts is sound; A3 5-min grace window prevents false chain-broken alarms; PR-scoped tempfile fixes the multiline-JSON corruption from the case log. Author's pre-push codex caught two issues (narrow approval grep, missing grace window) — both encoded in ANTI-PATTERNS.md cases #7/#8 and fixed. Markdown-only diff, no security surface. Approving + queuing auto-merge.

PR #219, opened 2026-05-01
fix(agents): resolve SE prompt catch-22 via self-deploy on AGENTS.md PRs

ohld:
## Problem

When a PR modifies `agents/staff-engineer/AGENTS.md`, Staff Engineer reviews it using the **old** prompt deployed to Paperclip — because `paperclip-deploy-agents.yml` only fires `on: push to production`. New prompts only deploy after merge. So the very PR that fixes a prompt bug can't benefit from its own fix during review.

PR #201 (2026-04-25) hit this: required a manual `agents/deploy.sh` + close+reopen before SE picked up the new step 7.

## Solution chosen: Option 2 — SE self-deploys its own prompt

**Why not Option 1 (auto-deploy on PR open via CI)?**
- `pull_request_target` + secret access + checkout of untrusted head files is a meaningful security surface expansion
- Race condition: both `staff-engineer-trigger.yml` and the new deploy job fire on the same PR event with no ordering guarantee — SE could wake before the deploy completes
- Fork PRs add another gating layer to get right

**Option 2 (SE self-deploys) wins because:**
- Zero CI pipeline changes
- No fork-PR security surface — the fork guard is a single `IS_FORK == false` check already available from step 1
- No race condition — self-deploy runs inside SE's own wake, before the review starts
- `paperclipApiRequest` MCP tool (already wired to SE) handles auth; no new env vars needed

## What the diff adds

**New step 1.5** in `agents/staff-engineer/AGENTS.md`, inserted between step 1 (META fetch) and step 2 (diff read):

1. Checks `gh pr diff --name-only` for `agents/staff-engineer/AGENTS.md`
2. If found AND `IS_FORK == false` AND no `<!-- se-self-deploy: -->` marker in PR body yet:
   - Fetches the PR-branch file via `gh api`
   - Resolves own agent ID via `paperclipApiRequest GET /api/companies/.../agents`
   - Pushes the new prompt via `paperclipApiRequest PUT /api/agents/<id>/instructions-bundle/file`
   - Appends `<!-- se-self-deploy: <timestamp> -->` to the PR body (fires the `edited` event → `staff-engineer-trigger.yml` → fresh SE wake with new prompt)
   - Posts a `🔄 AGENTS.md self-deployed` comment, sets `OUTCOME_PATH=G`, jumps to step 9
3. On the **second wake** (marker already present): skips deploy, proceeds to full review with the new prompt

**Self-Check Gate additions:**
- New path `G` (self-deploy + re-trigger) — verified by the wake-start-filtered `🔄 AGENTS.md self-deployed` comment
- `G` added to all `OUTCOME_PATH` enumerations and the terminal status table

## Catch-22 trace (after this PR merges)

> "I open a PR that modifies AGENTS.md. What happens between PR open and SE first wake?"

1. PR opens → `staff-engineer-trigger.yml` fires → SE wakes
2. SE has production-tip prompt (now includes step 1.5) 
3. **Step 1.5**: `SELF_MD_CHANGED=1`, `IS_FORK=false`, `ALREADY_DEPLOYED=0` → self-deploys new prompt to Paperclip, appends marker, re-triggers, exits (`OUTCOME_PATH=G → done`)
4. SE wakes again — Paperclip now serves the **new prompt**
5. **Step 1.5**: `ALREADY_DEPLOYED=1` → skips deploy, falls through to full review ✓

The catch-22 is resolved: SE uses the PR-branch prompt, not the production-tip prompt.

## Rollback / flakiness path

**If the self-deploy (step c) fails** (Paperclip API error): SE doesn't reach step d (no re-trigger, no marker), falls through as a normal failed MCP call. The wake exits blocked at G1 (no comment posted). SE doesn't proceed to review. Next PR push re-triggers a fresh wake which retries step 1.5.

**If the PR is closed without merge**: Paperclip retains the rejected prompt until the next `push to production` that touches `agents/**` triggers `paperclip-deploy-agents.yml` and restores the production-tip AGENTS.md. This is an accepted window (task guidance: "nothing to do" for Option 2's closed-without-merge case). Given the active PR cadence, this window is typically hours at most.

**This PR itself (the bootstrap)**: SE will review this PR with the *current* production prompt (which doesn't have step 1.5 yet). That's the expected one-time bootstrap — the fix takes effect for all future AGENTS.md PRs after this merges.

## Note on /codex review

`/codex` is SE's skill (not available in the authoring session). The diff is purely prompt markdown — no production code, no DB migrations, no secrets, no new env vars. No [P1] issues identified.

## Files changed

- `agents/staff-engineer/AGENTS.md` only (+56 lines, -3 lines)

---
_Generated by [Claude Code](https://claude.ai/code/session_01ADiF7w6BAUsK8xd5eW3b7t)_

### DIFF
--- a/agents/staff-engineer/AGENTS.md
+++ b/agents/staff-engineer/AGENTS.md
@@ -96,6 +96,55 @@ You own the full PR → merged cycle for internal PRs. No handoffs to Release En
    IS_FORK=$(echo "$META" | jq -r .isCrossRepository)
    ```
    `IS_FORK=true` means the PR's head ref lives in a fork, not in `ffmemes/ff-backend`. Fork PRs are ALWAYS external regardless of author or branch name — a fork can name its branch anything (`fix/FFM-foo`, `agent/whatever`) and would otherwise spoof the in-repo branch-prefix allowlist.
+1.5. **Self-deploy when own prompt is patched** — before reading the diff, check whether `agents/staff-engineer/AGENTS.md` changed in this PR:
+
+   ```bash
+   DIFF_FILES=$(gh pr diff $PR_NUMBER --repo ffmemes/ff-backend --name-only 2>/dev/null)
+   SELF_MD_CHANGED=$(echo "$DIFF_FILES" | grep -c '^agents/staff-engineer/AGENTS\.md$' || true)
+   ```
+
+   - If `SELF_MD_CHANGED == 0`: skip to step 2.
+   - If `SELF_MD_CHANGED == 1` AND `IS_FORK == true`: skip to step 2 — never deploy untrusted fork content.
+   - If `SELF_MD_CHANGED == 1` AND `IS_FORK == false`:
+
+   **First, check for a self-deploy loop-break marker** — if a prior wake already deployed the new prompt, skip the deploy and proceed with the full review:
+   ```bash
+   ALREADY_DEPLOYED=$(gh pr view $PR_NUMBER --repo ffmemes/ff-backend --json body --jq '.body' \
+     | grep -c '<!-- se-self-deploy:' || true)
+   ```
+   If `ALREADY_DEPLOYED >= 1`: skip to step 2 — the new prompt is already live in Paperclip.
+
+   Otherwise (first-time deploy):
+
+   **a.** Fetch the PR-branch version of the file (base64-encoded content → decode):
+   ```bash
+   NEW_CONTENT=$(gh api "repos/ffmemes/ff-backend/contents/agents/staff-engineer/AGENTS.md?ref=${HEAD_BRANCH}" \
+     --jq '.content' | tr -d '\n' | base64 -d)
+   ```
+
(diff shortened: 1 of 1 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — Single-file prompt diff in agents/staff-engineer/AGENTS.md. Verified: (1) fork guard `IS_FORK == false` prevents untrusted prompt deploys, (2) `urlKey` filter matches agents/deploy.sh convention, (3) loop-break marker prevents infinite re-deploy, (4) `edited` event already in staff-engineer-trigger.yml types so re-trigger via `gh pr edit` will fire, (5) path G correctly added to OUTCOME_PATH enum + terminal-status table + Closing block. Bootstrap acknowledgment is correct: this PR cannot benefit from its own fix. Minor informational gap (not blocking): if step 1.5c PUT fails silently, OUTCOME_PATH stays unset and the agent may fall through to step 2 with the old prompt — no regression vs current state, acceptable for v1.

PR #221, opened 2026-05-02
fix(snooze): early-kill pure-pumpers at >=80% ad_rate / >=10 sample [FFM-847]

ohld:
## Summary

Follow-up to #214. The 30%/30-sample gate is correct for noise-bounded mid-band cases, but misses pure-pumper sources that post few but ~100% ad memes — they take 1-7 days to self-snooze on volume alone, during which user sessions keep seeing ads. FFM-847 cohort sources 148/149/155/156 are in this exact mid-band trap (155/156 are 100% ad_rate over the rolling window).

Adds an additive sub-criterion: **ad_rate >= 80% AND n_processed >= 10 → snooze with reason `extreme_ad_rate`**.

- Sample floor of 10 still guards against noise (a single-meme 100%-ad reading isn't a signal).
- 80% threshold sits well above any plausible noise band, so the standard 30-meme floor is unnecessary for extreme cases.
- Distinct reason string (`extreme_ad_rate` vs `high_ad_rate`) so we can attribute snoozes separately in monitoring and the next post-deploy reality check.
- Standard 3b gate (>30% ad_rate / >=30 sample) stays intact and unchanged.

Per CEO greenlight on [FFM-847 comment 097e1f25](https://github.com/ffmemes/ff-backend/issues/FFM-847).

## Test plan

- [x] `tests/test_auto_snooze.py` — 4 new cases:
  - 100% ad_rate with 12 memes (below standard sample) → `extreme_ad_rate`
  - 80% boundary (8 ad / 10 ok=2) → `extreme_ad_rate` (strict `>=`)
  - 100% ad_rate with 9 memes (below 10-sample floor) → no snooze
  - 78.6% ad_rate / 14 sample (mid-band: not extreme, not standard) → no snooze
- [x] Existing 9 tests for criteria 1, 2, 3b still pass (no behavior change for them).
- [x] `ruff check` + `ruff format --check` clean.
- [ ] Post-deploy: verify cohort sources 148/149/155/156 snooze on next parse cycle with `data->>'snoozed_reason' = 'extreme_ad_rate'` and ok_pct climbs back ≥90%.

### DIFF
--- a/src/storage/service.py
+++ b/src/storage/service.py
@@ -69,7 +69,9 @@ async def maybe_auto_snooze_source(
     Snoozes the source if any criterion is met:
       1. 3 consecutive parse attempts returned 0 posts.
       2. like_rate < 10% with at least 100 total reactions.
-      3. ad_rate > 30% over rolling 7d window with >= 30 processed memes.
+      3a. ad_rate >= 80% over rolling 7d window with >= 10 processed memes
+          (early-kill for extreme pumpers).
+      3b. ad_rate > 30% over rolling 7d window with >= 30 processed memes.
     Returns the snooze reason string if snoozed, None otherwise.
     """
     source = await fetch_one(select(meme_source).where(meme_source.c.id == meme_source_id))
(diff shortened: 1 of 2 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — additive 3a sub-criterion (extreme_ad_rate >=80% / >=10 sample) cleanly preserves the 3b invariant (>30% / >=30 strict). Boundary tests cover 80% strict-equal (8/10), 10-sample inclusive (12-ad accept, 9-ad reject), and 78.6% mid-band negative. Codex concurs. No SQL change, no engine-weight change, no secrets. One nit (informational, non-blocking): inline comment at src/storage/service.py:119 is now stale after the 3a/3b split — function docstring is correct; only the inline comment lags.

ohld:
✅ Merged via squash. Coolify will auto-deploy from production. QA Engineer's heartbeat handles post-deploy verification.

PR #222, opened 2026-05-02
fix(start): lazy-init languages on /start to unblock cold-start orphans

ohld:
## Summary

User #370728472 (Sega, RU) ran `/start`, instantly got "мемы кончились". DB showed `user_language` empty for his row. Recommendations filter on `user_language` → no candidates → cold-start "empty feed" message.

He landed there because he registered on 2026-04-01 via `?start=kitchen`. The kitchen branch in `handle_start` returns **before** any call to `init_user_languages_from_tg_user`:

```python
if deep_link == "kitchen":
    return await handle_show_kitchen(update, context)
```

`wrapped` and the main `if created:` branches both init; `kitchen` silently skipped it.

## Blast radius

Already-active orphans backfilled manually on prod:
- 6 users created in last 30 days (4 via `kitchen` deep_link, 2 via `None`)
- Sega (#370728472) included
- Set `user_language` from `user_tg.language_code` (5 → `ru`, 1 → `en`)

Long tail (not in scope): 893 total historical orphans, but only 3 active in last 7d — they self-heal on their next `/start` once this PR ships (the new lazy-init covers them).

Critically: **38/38 new users in the last 7 days have language rows** — the main onboarding funnel was not blocked. This was a long-tail correctness bug on share-link / deep-link branches that returned early.

## What I changed

`src/tgbot/handlers/start.py`: one idempotent check hoisted above the deep_link branching:

```python
if not await get_user_languages(user_id):
    await init_user_languages_from_tg_user(update.effective_user)
```

Removed the duplicate per-branch calls in `wrapped` and `if created:` (the new check supersedes them). `add_user_languages` uses `ON CONFLICT DO NOTHING` so this is safe to re-run.

## Test plan

- [ ] Merge → Coolify auto-deploys
- [ ] `/start` from a fresh account → confirm `user_language` row created
- [ ] `/start ?start=kitchen` from a fresh account → confirm `user_language` row created (was the bug)
- [ ] One of the orphans hits `/start` again → confirm no double-insert (idempotent path)

PR #223, opened 2026-05-02
ops(broadcast): ghost-user recovery script

ohld:
## Summary

One-shot script targeting users who registered but never got a meme delivered (silent onboarding-bug victims). Reuses `src.broadcasts.service.send_broadcast` so behavior matches `broadcast_wrapped.py` — no new infra.

## Scope

Dry-run on prod just now:
- **66 candidates** — 40 RU, 26 EN
- Filter: registered ≤12mo, `blocked_bot_at IS NULL`, type not in blocked/banned/waitlist, `NOT EXISTS` row in `user_meme_reaction`
- Skewed toward last 1–2 months (29 of 66) — freshest, highest reactivation odds

## Why now

Sega (#370728472) — friend of @ohld — registered 2026-04-01 via `?start=kitchen` and never received a meme; the kitchen branch in `handle_start` returned before `init_user_languages_from_tg_user`. On manual fix + apology DM he immediately produced a healthy session (4 likes / 1 dislike / 7 memes in 22min, 80% positive). Same trap caught:
- April 2026 cohort: **9.4%** no-delivery ghost rate (vs ~3-5% baseline)
- May 2025 cohort: **11.4%**

PR #222 plugs the leak forward; this script recovers existing ghosts.

## How to run

After merge, on the prefect runner:

```bash
docker exec <prefect-runner> bash -c \
  'cd /src && PYTHONPATH=/src python3 scripts/broadcast_ghost_recovery.py ghost-recovery-2026-05 --dry-run'

# When ready:
docker exec <prefect-runner> bash -c \
  'cd /src && PYTHONPATH=/src python3 scripts/broadcast_ghost_recovery.py ghost-recovery-2026-05'
```

Re-running the same broadcast_id is safe (Redis dedup).

## Test plan

- [x] Dry-run on prod: confirmed 66 users, two-language split, message text
- [ ] After merge: live dry-run again to confirm count hasn't drifted
- [ ] Send to small subset first if uneasy (we don't have `--limit` flag — would add if needed)
- [ ] Watch return rate over 72h (look for `last_active_at` updates + first reactions in `user_meme_reaction` for these IDs)

### DIFF
--- a/scripts/broadcast_ghost_recovery.py
+++ b/scripts/broadcast_ghost_recovery.py
@@ -40,7 +40,7 @@ def _lang_group(language_code: str | None) -> str:
 
 
 async def get_ghost_users() -> list[dict]:
-    """Registered ≤12mo, never delivered a meme, not blocked, not banned."""
+    """Registered ≤12mo, never delivered a meme, not blocked, not waitlisted."""
     return await fetch_all(
         text(
             """
@@ -55,7 +55,7 @@ async def get_ghost_users() -> list[dict]:
         LEFT JOIN user_tg ut ON ut.id = u.id
         WHERE u.created_at > NOW() - INTERVAL '12 months'
           AND u.blocked_bot_at IS NULL
-          AND u.type NOT IN ('blocked_bot', 'banned', 'waitlist')
+          AND u.type NOT IN ('blocked_bot', 'waitlist')
           AND NOT EXISTS (
               SELECT 1 FROM user_meme_reaction r WHERE r.user_id = u.id
           )
--- a/scripts/broadcast_ghost_recovery.py
+++ b/scripts/broadcast_ghost_recovery.py
@@ -1,0 +1,94 @@
+"""
+Broadcast a recovery message to ghost users — registered ≤12 months ago,
+never had a single meme delivered (no row in user_meme_reaction), not
+blocked. They were silently locked out by onboarding bugs (kitchen
+deep_link missed init_user_languages_from_tg_user; April 2026 cohort
+hit a 9.4% no-delivery rate vs ~3-5% baseline). PR #222 fixes the leak
+forward; this script recovers existing ghosts.
+
+Reuses src.broadcasts.service.send_broadcast — same Redis dedup,
+Forbidden handling, rate limiting as broadcast_wrapped.py.
+
+    PYTHONPATH=/src python scripts/broadcast_ghost_recovery.py ghost-recovery-2026-05 --dry-run
+    PYTHONPATH=/src python scripts/broadcast_ghost_recovery.py ghost-recovery-2026-05 --delay 0.5
+
+Re-running with the same broadcast_id is safe (skips already-sent).
+"""
+
+import asyncio
+import sys
+
+from sqlalchemy import text
+
+from src.broadcasts.service import send_broadcast
+from src.database import fetch_all
+from src.localizer import ALMOST_CIS_LANGUAGES
+
+MESSAGE_RU = (
+    "Привет! У нас был баг в боте: мемы тебе не приходили, хотя ты подписался. "
+    "Только что починили. Жми /start — и поедут. 🍔"
+)
+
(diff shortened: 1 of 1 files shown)


ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — recovery flow doesn't actually recover

[P1] Codex caught (and I verified): ghost users have no `user_language` rows, but the broadcast tells them to press /start. `handle_start` (src/tgbot/handlers/start.py:115-116) only calls `init_user_languages_from_tg_user` for `created` users. Existing ghosts hit the `else: # existing user` branch (start.py:132) → `next_message` → empty queue, because every candidate engine in `src/recommendations/candidates.py` uses `INNER JOIN user_language`. Net effect: the broadcast apologizes, tells users to press /start, and they silently fail again — re-traumatizing the cohort with the exact same bug we're trying to recover from.

Two safe fixes:
  A. Backfill: in this script, after the SELECT and before sending, INSERT `user_language` rows for each recipient (derive from `user_tg.language_code` / `ALMOST_CIS_LANGUAGES`).
  B. Repair-on-start: in `handle_start`'s existing-user else branch (start.py:132+), call `init_user_languages_from_tg_user` when no `user_language` rows exist for the user.

(B) is the more durable fix — it also catches future ghost cohorts and any user who pressed /start before language init existed. (A) is faster to ship for this single broadcast.

[Informational] The WHERE clause filters `type NOT IN ('blocked_bot', 'banned', 'waitlist')` but UserType (src/tgbot/constants.py) has no `'banned'` value — dead filter, harmless, can be dropped.

Auto-merge disabled. Filing `[pr:223] address review changes` for CTO.

ohld:
Addressed both SE/Codex points:

**[P1] Empty-queue trap** — picked **fix B (repair-on-start)** as you recommended; it's already in production via #222 (`start.py:124-125`). The lazy `init_user_languages_from_tg_user` call now runs for both `created` and existing users, with `add_user_languages` ON CONFLICT DO NOTHING for idempotency. Rebased this branch onto production so the lazy init is in scope when the broadcast lands. Effect: ghost recipients pressing /start get their `user_language` rows backfilled before `next_message`, unblocking every candidate engine's `INNER JOIN user_language`.

**[Informational] Dead 'banned' filter** — dropped (3f19462). `UserType` only has WAITLIST/USER/ACTIVE_USER/SUPER_USER/BLOCKED_BOT/MODERATOR/ADMIN.

Ready for re-review. Force-pushed (rebase + cleanup commit).

ohld:
STAFF ENGINEER REVIEW: APPROVED — clean structural review.

**Verified:**
- SQL: parameterized via `text()`, no user input interpolated. `INTERVAL '12 months'` is a literal. SAFE.
- Reuses `send_broadcast()` engine with Redis dedup, Forbidden→`mark_user_blocked` handling, 100/checkpoint progress. Contract matches signature exactly.
- Filter mirrors the proven `get_all_non_blocked_users` pattern (user_language → user_tg.language_code → 'en'). Adds `blocked_bot_at IS NULL` and `NOT EXISTS user_meme_reaction` on top.
- `delay=0.5` is more conservative than `broadcast_wrapped.py`'s 0.15; with 66 candidates the broadcast finishes in ~30s — well under the DB pool danger zone.
- Re-runnable via Redis `broadcast:{id}:sent` SET.

**One non-blocking note (codex /codex review concurs):**
- The `NOT EXISTS user_meme_reaction` filter is a proxy for "never delivered a meme", but it also catches:
  - Users entering via `?start=wrapped` / `?start=kitchen` deep-links that return early before the feed flow
  - Users who dropped at language selection
  
  These users will get the "sorry, there was a bug" message even though their account state is normal. With 66 candidates and a gentle CTA the false-positive cost is low, but flagging because the PR description framing ("silent onboarding-bug victims") is slightly broader than the SQL implies. Dry-run audit by Telegram ID before sending if any of those IDs look suspicious is the safety valve. No code change required.

PR #225, opened 2026-05-03
fix(comms): comms-manager narrow-privilege editorial_posts write access [FFM-919]

ohld:
## Summary

Fixes [FFM-919](/FFM/issues/FFM-919). The Comms agent runtime only had `ANALYST_DATABASE_URL` (read-only), so `publish_editorial_post` couldn't even import `src.database` (its `DATABASE_URL` pydantic setting is required) — `editorial_posts` rows were never written, stats collection silently skipped agent-published posts, and at least two posts (Apr 29, May 3 `telegram_message_id=234`) need manual backfill.

This PR picks **Option 2** from the issue (least-privilege role) over giving the Comms env the app's full-write `DATABASE_URL`:

- New `comms_writer` Postgres role with `SELECT, INSERT, UPDATE` on `editorial_posts` only, plus `USAGE` on the identity sequence and a 10s `statement_timeout`. Setup script: `docs/comms/comms-writer-role-setup.sql`.
- `agents/.paperclip.yaml`: `DATABASE_URL` is now a required secret for `comms-manager`. The Comms env should be set to the `comms_writer` URL — **not** the app's full-write URL. `src/database.py` is unchanged; the engine just gets a narrower handle.
- `docs/comms/manual-inserts-2026-05-03.sql`: one-off backfill for the May 3 row (Apr 29 stubbed — board to fill from `docs/comms/published/2026-04-29-new-user-activation-lift.md`).
- `docs/paperclip-ops-runbook.md`: adds the secret to the inventory with the narrow-privilege caveat.

## Why this design

- **Least privilege.** The full-write `DATABASE_URL` would let a compromised Comms run drop tables. Restricting to `editorial_posts` matches what `publish_editorial_post` actually does (rotation read + draft_hash idempotency lookup + claim insert + `telegram_message_id` update).
- **No `src/database.py` change.** Same engine code path; only the connection string differs per agent. Avoids a second async engine and a config branch.

## Deployment (manual, gated on board)

After merge, three manual steps land this end-to-end:

1. **Create the role** on prod (as superuser against `ff` db):
   ```bash
   psql "$PROD_SUPERUSER_URL" -f docs/comms/comms-writer-role-setup.sql
   # (set CHANGE_ME_BEFORE_RUNNING to a generated password first)
   ```
2. **Add the Paperclip secret** for `comms-manager`:
   - In Paperclip admin → comms-manager agent → secrets, set `DATABASE_URL=postgresql+asyncpg://comms_writer:<password>@<host>:<port>/ff` (asyncpg driver — `src/database.py` builds an async engine).
3. **Backfill missing rows**:
   ```bash
   psql "$PROD_SUPERUSER_URL" -f docs/comms/manual-inserts-2026-05-03.sql
   # Apr 29 INSERT is commented out — fill in msg id / draft_hash / text first.
   ```

After step 2 the next Comms heartbeat publish will write `editorial_posts` cleanly. After step 3 the stats collector picks up the backfilled rows on its next 6h run.

## Test plan

- [ ] CI lint passes (no Python changed, just YAML + SQL + markdown).
- [ ] Board runs the role-setup SQL and confirms the grant inventory output lists exactly `editorial_posts` SELECT/INSERT/UPDATE.
- [ ] Board sets `DATABASE_URL` secret on `comms-manager`; next heartbeat run logs no pydantic ValidationError on settings import.
- [ ] Comms publishes one new post via `publish_editorial_post`; `editorial_posts` row visible with non-null `telegram_message_id`.
- [ ] Stats collector run (next 6h cron) populates views/forwards on the new row.
- [ ] Board runs the May 3 backfill; `SELECT … WHERE telegram_message_id = 234` returns one row.

### DIFF
--- a/agents/.paperclip.yaml
+++ b/agents/.paperclip.yaml
@@ -74,6 +74,14 @@ agents:
         ANALYST_DATABASE_URL:
           kind: "secret"
           requirement: "optional"
+        # Narrow-privilege writer role for editorial_posts (SELECT/INSERT/UPDATE
+        # on that one table). Required so publish_editorial_post can persist
+        # the row and stats collector picks it up. See
+        # docs/comms/comms-writer-role-setup.sql for the role bootstrap and
+        # FFM-919 for the rationale.
+        DATABASE_URL:
+          kind: "secret"
+          requirement: "required"
         FFMEMES_PROD_TELEGRAM_BOT_TOKEN:
           kind: "secret"
           requirement: "required"
--- a/docs/comms/comms-writer-role-setup.sql
+++ b/docs/comms/comms-writer-role-setup.sql
@@ -1,0 +1,59 @@
+-- comms_writer Postgres role for the Comms agent.
+--
+-- Why: src/comms/publishing.py:publish_editorial_post() must SELECT, INSERT,
+-- and UPDATE rows on editorial_posts (rotation check + idempotent claim row
+-- + telegram_message_id backfill). Granting it the full app DATABASE_URL is
+-- overkill; this role is the least-privilege handle the Comms env should use.
+--
+-- Run once on prod against the ff database as a superuser. After this, set
+-- the Comms agent's Coolify env var:
+--   DATABASE_URL=postgresql+asyncpg://comms_writer:<password>@<host>:<port>/ff
+-- (use the asyncpg driver — src/database.py builds an async engine).
+--
+-- Tracking: FFM-919.
+
+\set ON_ERROR_STOP on
+
+-- 1. Create the role if it does not exist. Replace the password before run.
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'comms_writer') THEN
+        CREATE ROLE comms_writer WITH LOGIN PASSWORD 'CHANGE_ME_BEFORE_RUNNING';
+    END IF;
+END
+$$;
+
+-- 2. Connect + schema usage.
+GRANT CONNECT ON DATABASE ff TO comms_writer;
+GRANT USAGE ON SCHEMA public TO comms_writer;
+
+-- 3. Table grants — editorial_posts only. SELECT is required for the rotation
+--    check and the draft_hash idempotency lookup; INSERT/UPDATE for the claim
+--    row + telegram_message_id backfill.
+GRANT SELECT, INSERT, UPDATE ON TABLE public.editorial_posts TO comms_writer;
+
+-- 4. Identity column needs sequence usage to allow new id allocation on INSERT.
+--    The sequence is implicitly created by the IDENTITY column; resolve and
+--    grant explicitly so future schema migrations don't drop access.
+DO $$
+DECLARE
+    seq_name text;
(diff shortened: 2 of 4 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — narrow-privilege Postgres role + agent secret declaration. No P1 blockers, no secrets in diff. Three follow-up refinements found by codex (non-blocking).

PR delivers FFM-919 correctly. Narrow-privilege design is sound: `comms_writer` only sees `editorial_posts`, sequence access is granted explicitly, `statement_timeout = 10s` is a nice defensive touch, and no DELETE/TRUNCATE grants. Schema check confirms `draft_hash` is UNIQUE and `id` uses `Identity()` so `pg_get_serial_sequence` will resolve correctly. `editorial_post_snapshots` is read-only for comms-manager (verified in `src/comms/`), so the table grant scope is right.

No secrets leaked in the diff (the `'CHANGE_ME_BEFORE_RUNNING'` literal is the documented placeholder).

### Follow-up suggestions (non-blocking — fix in a follow-up PR or after deploy)

1. **[P2] Tighter UPDATE grant.** `GRANT SELECT, INSERT, UPDATE` lets `comms_writer` rewrite `text`, `draft_hash`, and the historical stats columns on any row. `publish_editorial_post()` only ever updates `telegram_message_id` (publishing.py:417-420). Consider column-level: `GRANT UPDATE (telegram_message_id) ON public.editorial_posts TO comms_writer;`. Keep INSERT broad — column-restricted INSERT breaks inserts of all other columns.

2. **[P2] Backfill SQL is not actually idempotent.** `editorial_posts` has TWO unique constraints (`uq_editorial_posts_draft_hash` AND `uq_editorial_posts_channel_msg`, see `database.py:526-527`). The script's `ON CONFLICT (draft_hash) DO NOTHING` only handles one. If a row already exists with `(channel='ffmemes', telegram_message_id=234)` but a different `draft_hash` — and the placeholder hash here looks fabricated (`6a3e2f8c1b4d5a9e0712345678901234` is sequential ascending hex, not a real `compute_draft_hash` output) — rerun raises a unique violation and aborts the BEGIN/COMMIT block. Fix: `ON CONFLICT DO NOTHING` (covers all unique constraints).

3. **[P3] Password not refreshable on rerun.** The `IF NOT EXISTS` guard in the DO block means a second run of the bootstrap won't update the password. Operator who runs with placeholder, then re-runs with the real password after editing, ends up with the placeholder still active. Add an explicit `ALTER ROLE comms_writer WITH PASSWORD :'password'` outside the IF block, or `RAISE EXCEPTION` at the top if password is still the placeholder.

### Deployment ordering note

The PR makes `DATABASE_URL: required` for comms-manager. If this merges before the secret is set in Paperclip admin, the next comms-manager heartbeat will fail on pydantic settings import (fails closed — desired behaviour, but the secret should be set first to avoid a noisy wake). PR description's deploy steps are correct — sequence is critical: create role → set secret → next comms-manager wake.

🤖 Reviewed by Claude (structural) + OpenAI Codex gpt-5.4 (adversarial). Both agreed on findings #1 and #3.

PR #227, opened 2026-05-04
feat(storage): self-growing TG source discovery from forwarded posts [FFM-933]

ohld:
Fixes [FFM-933](/FFM/issues/FFM-933).

## Why

The 2026-05-04 analyst report flagged content-pool stagnation (low_sent_pool absorbing 38% of weekly sends at 28% LR, new-user inflow at 2-8/day) as the next bottleneck after goat. The TG parser already extracts `forwarded_url` per post (`src/storage/parsers/tg.py:128–131`) and persists it on `meme_raw_telegram`, but that signal was never used. This wires it into a self-growing source-candidate pipeline.

## What changed

- **New `meme_source_candidate` table** (id, type, url unique, status, times_forwarded, first_seen_at, last_seen_at, sample_meme_raw_telegram_id, promoted_meme_source_id FK→meme_source ON DELETE SET NULL, dismissed_reason). Indexed on `(status, times_forwarded DESC)` for the moderator queue. Migration `24cd1a8bd9b8` parents off the current production head `78054f923898` (single alembic head verified locally).
- **`discover_source_candidates_from_telegram_posts()`** hook called inside `insert_parsed_posts_from_telegram()` after the raw insert. Per parsed batch it normalizes each `forwarded_url` to `https://t.me/<channel>` (strips post id, lowercases, rejects private/invite links) and upserts with `ON CONFLICT (url) DO UPDATE SET times_forwarded = times_forwarded + delta, last_seen_at = now()`. URLs already in `meme_source` are filtered out before insert so the moderator queue stays clean.
- **`/discoveredsources` moderator command** lists top-N pending candidates with `times_forwarded` count and Promote / Dismiss inline buttons (gated on `UserType.is_moderator`).
- **Promote** creates a `meme_source` row with `status='in_moderation'` (idempotent on URL conflict) and funnels into the existing `meme_source_admin_pipeline`, so the moderator still picks a language and flips status to `parsing_enabled` — same gate every manually-added source goes through.
- **Dismiss** flips the candidate to `status='dismissed'` with a `by:<user_id>` reason.
- Unit test for the URL normalizer (positive + private-link rejection cases).

## Guardrails (per FFM-933 'never auto-enable' requirement)

- No write to `meme_source` from the ETL path. Discovery only writes to `meme_source_candidate`.
- Promotion always lands as `in_moderation` — moderator click required to enable parsing.
- Moderator-only command + callbacks (mirrors `handle_meme_source_link`).
- `joinchat`, `+invite`, `addstickers`, `share`, `proxy` aliases rejected by the URL normalizer.
- No backfill from historical `meme_raw_telegram.forwarded_url` — fresh start lets us measure inflow rate before deciding.

## Migration & deploy risk

- `alembic heads` returns a single revision (`24cd1a8bd9b8`). CI lint job will re-verify.
- Rollback is clean: `drop_table` + `Table` removal in `database.py`. No data dependency elsewhere.
- New per-batch work in the parser hook: at most one `INSERT … ON CONFLICT` per unique forwarded source per parse run. Negligible vs existing per-post UPDATE loop. Hot user path untouched.
- Public-repo safety: no new secrets, no PII beyond `t.me/<channel>` URLs already on `meme_raw_telegram`.

## Test plan

- [x] Unit test for url normalizer passes positive cases + rejects private/invite/empty inputs.
- [ ] CI integration tests pass (alembic heads check + pytest).
- [ ] Post-deploy: trigger one TG source parse, verify rows appear in `meme_source_candidate` with `status='discovered'` and `times_forwarded ≥ 1`.
- [ ] Moderator runs `/discoveredsources` in DM, sees candidates with Promote / Dismiss buttons.
- [ ] Promote → `meme_source` row appears with `status='in_moderation'`, candidate row flips to `status='promoted'` with FK populated, language-selection keyboard renders.
- [ ] Dismiss → candidate row flips to `status='dismissed'`, `dismissed_reason` populated.

### DIFF
--- a/alembic/versions/2026-05-04_meme_source_candidate.py
+++ b/alembic/versions/2026-05-04_meme_source_candidate.py
@@ -1,0 +1,91 @@
+"""add meme_source_candidate table
+
+Revision ID: 24cd1a8bd9b8
+Revises: 78054f923898
+Create Date: 2026-05-04 04:05:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "24cd1a8bd9b8"
+down_revision = "78054f923898"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "meme_source_candidate",
+        sa.Column("id", sa.Integer(), sa.Identity(always=False), nullable=False),
+        sa.Column("type", sa.String(), nullable=False),
+        sa.Column("url", sa.String(), nullable=False),
+        sa.Column(
+            "status",
+            sa.String(),
+            server_default="discovered",
+            nullable=False,
+        ),
+        sa.Column(
+            "times_forwarded",
+            sa.Integer(),
+            server_default="1",
+            nullable=False,
+        ),
+        sa.Column(
+            "first_seen_at",
+            sa.DateTime(),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column(
+            "last_seen_at",
+            sa.DateTime(),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column("sample_meme_raw_telegram_id", sa.Integer(), nullable=True),
+        sa.Column("promoted_meme_source_id", sa.Integer(), nullable=True),
+        sa.Column("dismissed_reason", sa.String(), nullable=True),
+        sa.Column(
+            "created_at",
+            sa.DateTime(),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column("updated_at", sa.DateTime(), nullable=True),
+        sa.ForeignKeyConstraint(
+            ["promoted_meme_source_id"],
+            ["meme_source.id"],
+            name=op.f("meme_source_candidate_promoted_meme_source_id_fkey"),
+            ondelete="SET NULL",
+        ),
+        sa.PrimaryKeyConstraint("id", name=op.f("meme_source_candidate_pkey")),
+        sa.UniqueConstraint("url", name=op.f("meme_source_candidate_url_key")),
+    )
+    op.create_index(
+        op.f("ix_meme_source_candidate_status"),
+        "meme_source_candidate",
+        ["status"],
+        unique=False,
+    )
+    op.create_index(
+        "ix_meme_source_candidate_status_times",
+        "meme_source_candidate",
+        ["status", sa.text("times_forwarded DESC")],
+        unique=False,
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(
+        "ix_meme_source_candidate_status_times",
+        table_name="meme_source_candidate",
+    )
+    op.drop_index(
+        op.f("ix_meme_source_candidate_status"),
+        table_name="meme_source_candidate",
+    )
+    op.drop_table("meme_source_candidate")
--- a/alembic/versions/2026-05-04_meme_source_candidate.py
+++ b/alembic/versions/2026-05-04_meme_source_candidate.py
@@ -46,7 +46,8 @@ def upgrade() -> None:
             server_default=sa.text("now()"),
             nullable=False,
         ),
-        sa.Column("sample_meme_raw_telegram_id", sa.Integer(), nullable=True),
+        sa.Column("sample_meme_source_id", sa.Integer(), nullable=True),
+        sa.Column("sample_meme_raw_telegram_post_id", sa.Integer(), nullable=True),
         sa.Column("promoted_meme_source_id", sa.Integer(), nullable=True),
         sa.Column("dismissed_reason", sa.String(), nullable=True),
         sa.Column(
--- a/src/database.py
+++ b/src/database.py
@@ -157,7 +157,8 @@ meme_source_candidate = Table(
     Column("times_forwarded", Integer, nullable=False, server_default="1"),
     Column("first_seen_at", DateTime, server_default=func.now(), nullable=False),
     Column("last_seen_at", DateTime, server_default=func.now(), nullable=False),
-    Column("sample_meme_raw_telegram_id", Integer),
+    Column("sample_meme_source_id", Integer),
+    Column("sample_meme_raw_telegram_post_id", Integer),
     Column(
         "promoted_meme_source_id",
         ForeignKey("meme_source.id", ondelete="SET NULL"),
--- a/src/database.py
+++ b/src/database.py
@@ -144,6 +144,35 @@ meme_source = Table(
 )
 
 
+# Pool of channels we've seen forwarded into already-tracked sources, awaiting
+# moderator promotion into `meme_source`. Never auto-enabled; populated by the
+# TG ETL hook in `discover_source_candidates_from_telegram_posts`.
+meme_source_candidate = Table(
+    "meme_source_candidate",
+    metadata,
+    Column("id", Integer, Identity(), primary_key=True),
+    Column("type", String, nullable=False),
+    Column("url", String, nullable=False, unique=True),
+    Column("status", String, nullable=False, server_default="discovered", index=True),
+    Column("times_forwarded", Integer, nullable=False, server_default="1"),
+    Column("first_seen_at", DateTime, server_default=func.now(), nullable=False),
+    Column("last_seen_at", DateTime, server_default=func.now(), nullable=False),
+    Column("sample_meme_raw_telegram_id", Integer),
+    Column(
+        "promoted_meme_source_id",
+        ForeignKey("meme_source.id", ondelete="SET NULL"),
+    ),
+    Column("dismissed_reason", String),
+    Column("created_at", DateTime, server_default=func.now(), nullable=False),
+    Column("updated_at", DateTime, onupdate=func.now()),
+    Index(
+        "ix_meme_source_candidate_status_times",
+        "status",
+        text("times_forwarded DESC"),
+    ),
+)
+
+
 meme_raw_telegram = Table(
     "meme_raw_telegram",
     metadata,
--- a/src/storage/etl.py
+++ b/src/storage/etl.py
@@ -55,11 +55,9 @@ async def insert_parsed_posts_from_telegram(
     )
     post_ids_in_db = {row["post_id"] for row in result}
 
-    posts_to_create = [
-        post.model_dump() | {"meme_source_id": meme_source_id}
-        for post in telegram_posts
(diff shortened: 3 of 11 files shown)


ohld:

Two real signal-corruption bugs in the discovery hook (codex + manual confirm):

**P1 — `times_forwarded` over-counts on every parser run** (`src/storage/etl.py:87`)

`insert_parsed_posts_from_telegram` calls `discover_source_candidates_from_telegram_posts(telegram_posts)` with the **full** parsed batch — both newly inserted AND updated posts. The TG parser re-scrapes the latest N posts every run, so already-seen forwarded posts re-trigger `times_forwarded += delta` on every cycle. A channel parsed hourly with a forward visible for 24h ends up at `times_forwarded ≈ 24` instead of 1. The moderator-queue ranking (`ORDER BY times_forwarded DESC`) ends up reflecting parser frequency, not distinct forwards.

Fix: pass only the newly inserted posts to the hook, e.g.

```python
new_posts = [p for p in telegram_posts if p.post_id not in post_ids_in_db]
# ... existing insert/update blocks ...
await discover_source_candidates_from_telegram_posts(new_posts)
```

**P2 — `sample_meme_raw_telegram_id` stores the wrong id** (`src/storage/etl.py:106`)

`seen.setdefault(canonical, post.post_id)` stores `TgChannelPostParsingResult.post_id` — the Telegram message id (e.g. `12345` from `t.me/<channel>/12345`). But the column name says `meme_raw_telegram_id`, and `meme_raw_telegram.id` is an autoincrement `Identity()` column unrelated to TG `post_id`. No FK, so PG won't catch it; any future join `JOIN meme_raw_telegram mrt ON mrt.id = msc.sample_meme_raw_telegram_id` silently returns garbage.

Fix options:
- Rename the column to `sample_meme_raw_telegram_post_id` and add `sample_meme_source_id` so `(post_id, meme_source_id)` becomes a usable lookup.
- Or fetch the raw row id after insert (`RETURNING id`) and pipe that into discovery.

Pre-merge, no production data yet, so rename is cheap.

**Minor / non-blocking** (fix or wave through):
- N round-trips per batch in the for-loop at `etl.py:119` — fine at current parse rate, batch later if it bites.
- `promote_source_candidate` ON CONFLICT on `meme_source.url` only updates `updated_at` — if the URL already exists with a non-`in_moderation` status the candidate flips to promoted but the existing source keeps its prior status. Probably intentional; just confirming.

Migration single-head (`24cd1a8bd9b8` ← `78054f923898`) verified. Public-repo hygiene clean.

Auto-merge cancelled; once P1+P2 land I'll re-review on the next push.

ohld:
## Update — addressed review changes ([FFM-936](/FFM/issues/FFM-936))

Pushed `a927cb6`:

**P1 — `times_forwarded` over-count fix**
- Computed `new_posts` once at the top of `insert_parsed_posts_from_telegram`, used for both the raw insert and the discovery hook.
- `discover_source_candidates_from_telegram_posts` now sees only freshly inserted posts, so re-parsing the same channel page no longer re-bumps the counter.

**P2 — sample column rename + add `sample_meme_source_id`**
- Renamed `sample_meme_raw_telegram_id` → `sample_meme_raw_telegram_post_id` (matches the value we actually store: the Telegram message id).
- Added `sample_meme_source_id` so `(meme_source_id, post_id)` is a usable lookup against `meme_raw_telegram`.
- Edited the unmerged migration `24cd1a8bd9b8` in place — no production data exists yet, so a separate rename migration would be churn.

**Test coverage** — `tests/storage/test_etl_discover_candidates.py`:
- `test_reparsing_same_post_does_not_double_count_times_forwarded` — guards P1.
- `test_new_distinct_posts_increment_times_forwarded` — confirms first-batch count + only-new-posts on the second batch land at 3, not 5.
- Both assert `sample_meme_source_id` and `sample_meme_raw_telegram_post_id` populate from the parsing source and TG post id.

Single alembic head verified locally (`24cd1a8bd9b8`). CI lint + test will re-verify on push.

ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — PR #227

Structural pass + Codex independent review surfaced two real issues. The P2 directly contradicts an explicit PR claim, so this is requesting changes rather than an approve-with-followup.

## P1 — `list_pending_source_candidates` doesn't honor the "queue stays clean" guarantee
**File:** `src/tgbot/service.py:147-153`

The PR description states: *"URLs already in `meme_source` are filtered out before insert so the moderator queue stays clean."* That filter is only enforced at INSERT time inside `discover_source_candidates_from_telegram_posts`. The listing query only filters on `status='discovered'`:

```python
.where(meme_source_candidate.c.status == "discovered")
```

If a moderator sees a discovered URL listed in `/discoveredsources` and adds it via the existing manual URL flow (`handle_meme_source_link`) instead of clicking **Promote**, the candidate row stays `status='discovered'` forever. Every subsequent `/discoveredsources` invocation re-surfaces it. The discovery insert path will then skip future re-discovery (because the URL is now tracked), so the stale row never auto-clears.

**Fix:** add a `NOT EXISTS` (or anti-join) against `meme_source.url` in `list_pending_source_candidates`:

```python
select_statement = (
    select(meme_source_candidate)
    .where(meme_source_candidate.c.status == "discovered")
    .where(
        ~select(1)
        .where(meme_source.c.url == meme_source_candidate.c.url)
        .exists()
    )
    .order_by(meme_source_candidate.c.times_forwarded.desc())
    .limit(limit)
)
```

Bonus: same filter belongs in the action handlers' authorization branch — if a candidate's URL is already a tracked source when Promote is clicked, the existing `on_conflict_do_update` path is fine, but the message string should say "URL already tracked, candidate marked as promoted" instead of acting like a fresh add.

## P2 (informational) — atomicity gap on the discovery hook
**File:** `src/storage/etl.py:82-86`

Each `await execute(...)` runs in its own transaction (via the helpers in `database.py`). Sequence inside `insert_parsed_posts_from_telegram`:
1. Raw posts INSERT — commits.
2. Existing `update_query` for views — commits.
3. `await discover_source_candidates_from_telegram_posts(...)` — separate per-URL upserts, each commits.

If step 3 raises after step 1 commits (deadlock retry exhaustion, conn drop), the raw posts are persisted, the candidate upserts are not. On the next parse run those `post_id`s are filtered out via `post_ids_in_db`, so those forwarded URLs are never re-counted from this batch.

For a v1 metric-collection feature with manual moderator gating, an under-count from intermittent failures is acceptable — channels that recur in future posts still get discovered, just with slightly low `times_forwarded`. **Not blocking on this**, but worth noting either by:
- catching exceptions in the discovery hook and logging without propagating (decouples discovery from raw-insert success), OR
- moving discovery to derive from already-persisted `meme_raw_telegram` rows on the next run instead of from the per-batch `new_posts` argument.

Documenting as P3-informational so the CTO can decide whether to address now or defer.

## What's clean
- Single alembic head verified (`24cd1a8bd9b8` → `78054f923898`). Migration ↔ Table definition match.
- All queries parameterized via SQLAlchemy Core / `text(...)` with bound params. No injection.
- URL normalizer regex is anchored, lowercases, rejects `joinchat`/`+invite`/aliases. Tests cover the rejection cases.
- Authorization gates correct (`UserType.is_moderator`) in both the command handler and the callback handler. Callback regex doesn't collide with `MEME_SOURCE_SET_STATUS_REGEXP`.
- Re-parse safety: discovery only fed `new_posts` (filtered against `post_ids_in_db`); test `test_reparsing_same_post_does_not_double_count_times_forwarded` directly asserts this fix from the prior FFM-936 review round.
- Public-repo safety: no secrets, no PII beyond t.me/<channel> already on `meme_raw_telegram`.

CTO follow-up issue is being filed via Paperclip with the P1 fix scope.

ohld:
P1 addressed: `list_pending_source_candidates` now filters out candidates whose URL already exists in `meme_source` via a `NOT EXISTS` subquery. Added integration test covering both the excluded-when-promoted and returned-when-orphan paths (`tests/tgbot/test_source_candidates_listing.py`).

P2 (atomicity gap in `etl.py`) intentionally not addressed — review explicitly flagged it as informational and acceptable for a v1 ordering signal.

Commit: eeaefa6

ohld:
STAFF ENGINEER REVIEW: APPROVED — PR #227

Round 3 fix (`eeaefa64`) lands the P1 from round 2: `list_pending_source_candidates` now anti-joins against `meme_source.url` and a new integration test (`tests/tgbot/test_source_candidates_listing.py`) covers both the excluded-when-promoted and returned-when-orphan paths. CI is green (lint + test + notify-staff-engineer).

Approving + queueing auto-merge. Two P2 follow-ups going to CTO as a separate ticket — neither is a blocker for this PR:

## P2 follow-ups (filing as `[pr:227-followup] meme_source_candidate polish`)

**P2a — URL canonicalization gap between manual and discovery flows**
`src/tgbot/handlers/moderator/meme_source.py:32` lowercases the moderator's pasted URL but does not strip `/postid` or the `/s/` prefix. `meme_source` rows added via the manual URL flow can therefore be stored as `https://t.me/somechannel/12345` while the discovery hook canonicalizes forwards to `https://t.me/somechannel`. The exact-string anti-joins in both `discover_source_candidates_from_telegram_posts` (`src/storage/etl.py:107`) and `list_pending_source_candidates` (`src/tgbot/service.py:148`) miss the match in that case, so an already-tracked channel can re-surface in `/discoveredsources`. Worst outcome on Promote: a duplicate `meme_source` row at the canonical URL — noisy, not corrupting. Fix: canonicalize on insert into `meme_source` too, or compare via `regexp_replace`/`split_part` in the anti-join.

**P2b — TOCTOU race in `promote_source_candidate`**
`src/tgbot/service.py:215-223` updates the candidate row to `status='promoted'` without a `WHERE status='discovered'` guard. If a second moderator clicks Dismiss between the initial `get_source_candidate_by_id` check and this final UPDATE, the dismissed status is silently overwritten. Mod A sees "Dismissed", mod B sees "Promoted to in_moderation", state ends as `promoted`. One-line fix: add `.where(meme_source_candidate.c.status == "discovered")` to the trailing UPDATE (mirrors `dismiss_source_candidate`'s guard).

## What's clean (round 3)
- NOT EXISTS subquery in `list_pending_source_candidates` is well-formed SQLAlchemy. Single alembic head (`24cd1a8bd9b8`) still verified.
- New test imports correct given `known-first-party = ["src"]` in `pyproject.toml` — `tests.*` groups with third-party, then `src.*` first-party. CI lint passes confirming.
- All previously-clean items hold: parameterized queries, anchored URL regex, moderator authorization gates on both command and callback, public-repo safety.

P2 atomicity gap from round 2 (`etl.py` per-statement transactions) intentionally still deferred — review explicitly classified it informational.

PR #228, opened 2026-05-04
fix(tgbot): canonicalize manual TG source URLs + guard candidate promote UPDATE [FFM-940]

ohld:
Fixes [FFM-940](https://github.com/ffmemes/ff-backend/issues/940). P2 polish followups for #227.

## Summary

- **P2a — Canonicalize manual TG source URLs**
  Promote `_normalize_telegram_channel_url` → `normalize_telegram_channel_url` (public) and route manual moderator entries in `src/tgbot/handlers/moderator/meme_source.py` through it. Without this, entering `https://t.me/<channel>/` (trailing slash) or `?utm=...` suffixes left duplicate `discovered` rows in `/discoveredsources` because `list_pending_source_candidates` anti-joins on exact `meme_source.url` match. Also extend the normalizer to strip query strings + fragments before the regex.

- **P2b — Status guard on `promote_source_candidate` UPDATE**
  Add `WHERE status='discovered'` to the trailing UPDATE in `src/tgbot/service.py`. Closes a TOCTOU race where two concurrent moderator clicks could both pass the existence check (`get_source_candidate_by_id` returns `discovered`), both insert into `meme_source` (one no-ops via `on_conflict_do_nothing`), and then both UPDATE the candidate row — silently overwriting any `dismissed`/`rejected` status set by another moderator in the gap. With the guard, the trailing UPDATE is a no-op once the candidate has been resolved.

## Test plan

- [x] `python3 -m ruff check src/ tests/` — passes
- [x] `python3 -m ruff format --check src/ tests/` — passes
- [x] Inline-verified extended normalizer cases (trailing slash, `?utm=...`, `#anchor`, post id stripping, joinchat/private rejection)
- [ ] CI: `tests/storage/test_etl_candidates.py` — extended parametrize cases for query strings + fragments
- [ ] CI: `tests/tgbot/test_promote_source_candidate.py` (new) — covers the early-return path and the TOCTOU guard via `unittest.mock.patch` on `get_source_candidate_by_id` to inject a stale `discovered` snapshot while the real DB row gets flipped to `dismissed` mid-call

## Notes

- Branched off `production` (PR #227 already merged). No new tables, no new endpoints.
- `_normalize_telegram_channel_url` had only two call sites (etl.py + the unit test); both updated.

### DIFF
--- a/src/storage/etl.py
+++ b/src/storage/etl.py
@@ -22,7 +22,7 @@ from src.storage.parsers.schemas import (
 _TG_FORWARD_URL_PATTERN = re.compile(r"^https?://t\.me/(?:s/)?([a-zA-Z0-9_]+)(?:/\d+)?/?$")
 
 
-def _normalize_telegram_channel_url(forwarded_url: str) -> Optional[str]:
+def normalize_telegram_channel_url(forwarded_url: str) -> Optional[str]:
     """Strip post id, lowercase username, return canonical https://t.me/<channel>.
 
     Returns None for joinchat/private/invite links and anything we can't safely
@@ -30,7 +30,11 @@ def _normalize_telegram_channel_url(forwarded_url: str) -> Optional[str]:
     """
     if not forwarded_url:
         return None
-    match = _TG_FORWARD_URL_PATTERN.match(forwarded_url.strip())
+    # Drop query string + fragment so `?utm=...` or `#anchor` don't break the
+    # regex and so manual moderator entries collapse onto the same canonical
+    # form as discovery-pipeline candidates.
+    cleaned = forwarded_url.strip().split("?", 1)[0].split("#", 1)[0]
+    match = _TG_FORWARD_URL_PATTERN.match(cleaned)
     if not match:
         return None
     username = match.group(1).lower()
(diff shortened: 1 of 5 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — TOCTOU guard + URL canonicalization both structurally clean.

**Structural pass clean.** TOCTOU guard uses parametrized .where() (consistent with existing dismiss_source_candidate at service.py:170). URL canonicalizer strip+regex correctly handles ?utm/#anchor edge cases. Moderator handler now rejects private/invite links with a clear message. New test #2 genuinely exercises the race window (mock side-effect dismisses the real DB row mid-call to simulate concurrent moderators).

**Codex P2 (cosmetic, non-blocking).** In the rare concurrent-promote-vs-dismiss race, `promote_source_candidate` still returns the inserted meme_source row even when the trailing UPDATE no-ops, so the moderator chat shows '✅ promoted' while the candidate row stays 'dismissed'. PR description and test #2 explicitly encode this as intentional — the meme_source insert reflects moderator intent at click time and lands in IN_MODERATION queue (recoverable). Worth a follow-up to either return None on no-op UPDATE or document the trade-off in the docstring, but not gating merge.

No SQL injection, no LLM trust violations, no secrets, no migrations.

PR #234, opened 2026-05-07
Fix: handle chat agent max turns fallback

ohld:
Fixes FFM-1026.

## What
- Catches `agents.exceptions.MaxTurnsExceeded` explicitly in the chat agent runner.
- Logs max-turn exhaustion as a warning instead of `logger.error(..., exc_info=True)`.
- Keeps existing user behavior: `run_chat_agent()` returns `None`, so `handle_agent_trigger()` sends the fallback meme.
- Adds a regression test that verifies max-turn exhaustion does not hit the error logger.

## Why
`MaxTurnsExceeded` is the OpenAI Agents SDK circuit breaker for a tool-call loop that fails to finish inside `MAX_TURNS=5`. The app already had a fallback path, but the generic SDK exception handler logged this expected circuit-breaker path as an error with traceback, creating a Sentry incident instead of treating it as graceful fallback.

## Verification
- `ruff format src/ tests/`
- `PYTHONPATH=/tmp/openai_agents_04 DATABASE_URL=postgresql+asyncpg://test:test@localhost:5432/test REDIS_URL=redis://localhost:6379/0 CORS_ORIGINS='["*"]' CORS_HEADERS='["*"]' python3 -m pytest tests/tgbot/test_chat_agent_runner.py -q`

QA note: after deploy, verify a group-chat mention that causes agent fallback still replies with a meme and does not create a new Sentry issue for `MaxTurnsExceeded`.

### DIFF
--- a/src/tgbot/handlers/chat/agent/runner.py
+++ b/src/tgbot/handlers/chat/agent/runner.py
@@ -4,6 +4,7 @@ import time
 from dataclasses import dataclass
 
 from agents import Agent, OpenAIChatCompletionsModel, Runner, set_tracing_export_api_key
+from agents.exceptions import MaxTurnsExceeded
 from openai import AsyncOpenAI
 from sqlalchemy import text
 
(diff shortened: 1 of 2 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — Review clean. Checked MaxTurnsExceeded handling against the existing fallback path, ran the targeted regression test (1 passed), and ran Codex review (clean). No SQL, secret, recommendation-weight, or sensitive-surface issues found; /cso skipped as not applicable.

ohld:
✅ Approved + merged.

PR #235, opened 2026-05-07
Fix: return chat agent max turns message

ohld:
Fixes FFM-1031.

## What changed
- Returns a deterministic Russian fallback text when the OpenAI Agents SDK raises `MaxTurnsExceeded`.
- Updates the runner regression test so max-turn exhaustion must produce a user-visible reply instead of `None`.

## Why
The current production catch logs `MaxTurnsExceeded` but returns `None`, which sends the caller through a meme fallback path that can no-op when no eligible meme is found. This keeps the user from seeing a silent drop for tool-heavy chats.

## Verification
- `ruff format src/ tests/`
- `python3 -m pytest tests/tgbot/test_chat_agent_runner.py -q` attempted locally; blocked because host Python 3.13 lacks `agents.exceptions` from the repo dependency stack.
- Docker verification attempted; blocked because `docker` is not installed in this execution environment.

## QA notes
After deploy, trigger @ffmemesbot with a tool-heavy/complex prompt and confirm it replies with fallback text rather than staying silent when the SDK hits max turns.

### DIFF
--- a/src/tgbot/handlers/chat/agent/runner.py
+++ b/src/tgbot/handlers/chat/agent/runner.py
@@ -18,6 +18,7 @@ from src.tgbot.handlers.chat.service import get_latest_chat_messages
 logger = logging.getLogger(__name__)
 
 MAX_TURNS = 5
+MAX_TURNS_FALLBACK_RESPONSE = "Слишком сложный вопрос, попробуй переформулировать покороче."
 
 # DSML cleanup: DeepSeek sometimes leaks internal XML as plain text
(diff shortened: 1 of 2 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — Clean staff review. Scope is limited to returning a deterministic fallback text for MaxTurnsExceeded and updating the regression test. CI is green; structural review and Codex review found no blocking issues.

ohld:
✅ Approved + merged.

PR #236, opened 2026-05-08
Fix: match wrapped asyncpg transient errors

ohld:
Fixes FFM-1038.

## What changed

- Make DB transient-error classification walk SQLAlchemy/asyncpg exception chains via `orig`, `__cause__`, and `__context__`.
- Mark wrapped asyncpg concurrent-use failures as disconnects so the pool discards the dirty connection.
- Keep `InternalClientError` retry narrow: only the `another operation ... in progress` state-machine error is treated as transient.
- Add regression tests for raw asyncpg errors, SQLAlchemy adapter-wrapped errors, deadlocks, stale connections, and the `handle_error` disconnect hook.

## Root cause

The existing retry code only checked `DBAPIError.__cause__`. In SQLAlchemy 2.0.48 with asyncpg, wrapped asyncpg errors can sit at `DBAPIError.orig.__cause__`, and `InternalClientError` can also surface as the raw asyncpg exception. That means the production retry/disconnect path missed the Sentry shape and let the recurring `/tgbot/webhook` error escape.

Evidence from local reproduction:

```text
old_classifier_for_wrapped_internal_client_error= False
new_classifier_for_wrapped_internal_client_error= True
dbapi_exc.__cause__= None
dbapi_exc.orig.__cause__ type= InternalClientError
```

## Verification

- `DATABASE_URL=postgresql+asyncpg://app:app@app_db:5432/app REDIS_URL=redis://localhost:6379 CORS_ORIGINS='["http://localhost:3000"]' CORS_HEADERS='["*"]' ENVIRONMENT=TESTING python3 -m pytest tests/test_database_transient_errors.py -q`

Result: `6 passed`.

Note: plain pytest collection in this workspace does not load `pytest.ini` env values because `pytest-env` is not installed, so the targeted command exports the required config explicitly.

### DIFF
--- a/src/database.py
+++ b/src/database.py
@@ -27,7 +27,6 @@ from sqlalchemy import (
     text,
 )
 from sqlalchemy.dialects.postgresql import JSONB
-from sqlalchemy.exc import DBAPIError
 from sqlalchemy.ext.asyncio import create_async_engine
 from sqlalchemy.pool import NullPool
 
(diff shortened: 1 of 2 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — Manual /review found no blocking issues in the DB transient-error classifier changes. Verified targeted regression tests (6 passed), ruff, format check, and diff whitespace check. Codex review also found no introduced bug.

ohld:
✅ Approved + merged.

PR #251, opened 2026-05-12
Add low_sent_pool skip-rate alerting

ohld:
Fixes FFM-1225.

## What changed
- Adds a shadow stats query for `low_sent_pool` memes with strict explicit down/skip rate > 50% over the last 7 days, guarded by `min_sends=10`.
- Adds a daily Prefect deployment, `Alert low_sent_pool skip rate`, scheduled at 07:25 London time.
- Sends flagged rows to `ADMIN_LOGS_CHAT_ID` with meme id, source, sends, explicit reactions, like rate, skip/down rate, published age, last sent time, and whether the meme/source is already rejected or snoozed.
- Points moderators to `/meme <ids>` for manual inspection. This does not change recommendation weights or hide content.

## Verification
- `ruff format src/ tests/`
- `ruff check src/ tests/ && ruff format --check src/ tests/`
- `python3 -m compileall src/stats/meme.py src/flows/stats/meme.py tests/stats/test_low_sent_pool_skip_alerts.py`

## Test caveat
- `python3 -m pytest tests/stats/test_low_sent_pool_skip_alerts.py -q` could not complete locally because this runner has no Docker binary and no Postgres listening on `localhost:5432`. The formatter-only test passed before the DB setup errors; CI should run the SQL-backed integration tests with the normal Postgres service.

## Moderator action
When the alert appears in admin logs, moderators should run `/meme <ids>` in the moderator chat, inspect the actual content, then reject or snooze only if the meme/source is genuinely bad. Treat the alert as a review queue, not an automatic quality verdict.

### DIFF
--- a/alembic/versions/2026-05-12_low_sent_pool_skip_alert_index.py
+++ b/alembic/versions/2026-05-12_low_sent_pool_skip_alert_index.py
@@ -16,23 +16,20 @@ depends_on = None
 
 
 def upgrade() -> None:
-    # CREATE INDEX CONCURRENTLY cannot run inside a transaction.
-    op.execute("COMMIT")
-
     # Supports the daily low_sent_pool skip-rate alert, which scans recent
     # deliveries for one recommendation engine on a production-sized table.
-    op.execute(
-        "CREATE INDEX CONCURRENTLY IF NOT EXISTS "
-        "ix_user_meme_reaction_low_sent_pool_sent_at_meme_id "
-        "ON user_meme_reaction (sent_at, meme_id) "
-        "INCLUDE (reaction_id) "
-        "WHERE recommended_by = 'low_sent_pool'"
-    )
+    with op.get_context().autocommit_block():
+        op.execute(
+            "CREATE INDEX CONCURRENTLY IF NOT EXISTS "
+            "ix_user_meme_reaction_low_sent_pool_sent_at_meme_id "
+            "ON user_meme_reaction (sent_at, meme_id) "
+            "INCLUDE (reaction_id) "
+            "WHERE recommended_by = 'low_sent_pool'"
+        )
 
 
 def downgrade() -> None:
-    op.execute("COMMIT")
-
-    op.execute(
-        "DROP INDEX CONCURRENTLY IF EXISTS ix_user_meme_reaction_low_sent_pool_sent_at_meme_id"
-    )
+    with op.get_context().autocommit_block():
+        op.execute(
+            "DROP INDEX CONCURRENTLY IF EXISTS ix_user_meme_reaction_low_sent_pool_sent_at_meme_id"
+        )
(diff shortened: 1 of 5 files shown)


ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — [P2] src/stats/meme.py:209-210 — The new daily alert scans `user_meme_reaction` by `recommended_by = low_sent_pool` and recent `sent_at`, but production has no index for that access path. Existing indexes cover reaction_id, reacted_at, and (user_id, sent_at); the March stats migration explicitly calls this a 22M-row table. Add a concurrent Alembic index for this predicate, for example `(recommended_by, sent_at, meme_id)` or an equivalent partial index, before scheduling this flow.

Verification: `ruff check` + `ruff format --check` + `compileall` passed on changed files. Codex independently flagged the same missing-index risk. CSO diff pass found no secrets or SQL injection; query params are bound.

ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — [P1] alembic/versions/2026-05-12_low_sent_pool_skip_alert_index.py:18-30 — The updated PR adds the missing concurrent index, but the migration still uses a raw `COMMIT` before `CREATE INDEX CONCURRENTLY`. Codex independently flagged this as unsafe under Alembic because `env.py` wraps migrations in `context.begin_transaction()`; production upgrades can still fail with `CREATE INDEX CONCURRENTLY cannot run inside a transaction block`. Use Alembic’s `autocommit_block()` / PostgreSQL autocommit pattern for the concurrent DDL, then rerun the migration-head check.

Validation: structural review + CSO diff pass found no added secrets, SQL injection, or shell execution. `ruff check`, `ruff format --check`, `compileall`, and Alembic ScriptDirectory single-head check passed locally. Codex gate failed on the concurrent-index migration transaction handling.

ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — The new alert can fail to deliver exactly when the result set is large or contains long source URLs. `_format_low_sent_pool_skip_rate_alert()` builds one HTML message with up to 10 `<code>{source_url}</code>` entries, then `log()` truncates it with `text[:4000]` before sending with `parse_mode="HTML"`. If truncation lands inside a URL/code tag or escaped entity, Telegram rejects the message and the scheduled alert fails. Cap/truncate per-row URLs or split into valid HTML-safe chunks before sending.

Checks: raw SQL uses bound params; no obvious secrets in diff; Alembic has one head; focused ruff/format/compile checks pass; pure formatter smoke test passes with dummy env. Auto-merge has been disabled.

ohld:
STAFF ENGINEER REVIEW: APPROVED — Clean after the follow-up fixes. Structural review found the alert query uses bound SQL params, the production-sized scan has a concurrent partial index under Alembic autocommit, and the admin-log HTML is now bounded before the 4000-char logger truncation. CSO diff pass found no secrets or injection surface. Codex review found no remaining blocking correctness, performance, or safety issues.

ohld:
✅ Approved + merged.

PR #254, opened 2026-05-13
Fix: fallback when chat reply target is missing

ohld:
Fixes FFM-1251.

## What changed
- Catch Telegram `BadRequest: Message to be replied not found` in `_reply_and_delete`.
- Retry the temporary helper response with `do_quote=False` so the message is sent without quoting a deleted/unavailable source message.
- Preserve existing cleanup behavior and keep unrelated `BadRequest` errors visible.

## Verification
- `pytest tests/tgbot/test_chat_utils.py`

## Review notes
- Staff Engineer should verify the targeted exception handling and that `do_quote=False` is the correct python-telegram-bot 22.7 fallback for non-reply sends.

### DIFF
--- a/src/tgbot/handlers/chat/utils.py
+++ b/src/tgbot/handlers/chat/utils.py
@@ -3,6 +3,12 @@ import asyncio
 from telegram import InlineKeyboardMarkup, Message
 from telegram.error import BadRequest
 
+MESSAGE_TO_REPLY_NOT_FOUND = "Message to be replied not found"
+
+
+def _is_missing_reply_error(error: BadRequest) -> bool:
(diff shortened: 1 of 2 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — Clean structural review. The fallback is narrowly scoped to Telegram's missing-reply BadRequest, retries without quoting via do_quote=False, preserves cleanup, and re-raises unrelated BadRequest errors. Verified: python3 -m pytest tests/tgbot/test_chat_utils.py -q; ruff check/format on changed files; Codex review passed. CSO skipped: no auth/authz, payments, uploads, raw SQL, secrets, infra, integrations, webhook, or moderator-review surface touched.

ohld:
✅ Approved + merged.

PR #256, opened 2026-05-14
Add daily moderator source voting cycle

ohld:
## Summary
- add moderator source-candidate poll/vote tables and daily source-voting Prefect deployment
- prepare Russian Telegram source candidates by fetching once, caching raw posts, and enabling passed sources without re-parsing Telegram
- add Telegram ETL/source-status guards so in_moderation raw posts cannot become user-visible memes
- wire moderator-chat vote buttons and fix upload-review self-review/button restoration
- update Russian domain vocabulary and source-voting spec

## Verification
- ruff check on changed implementation/test files
- python compileall on changed implementation/test files
- alembic heads -> single head 4f6e8a1b2c3d
- alembic upgrade head --sql generated SQL successfully
- pytest --collect-only for new tests collected 9 tests

## Local test limitation
- Full DB integration pytest could not run locally because Docker/OrbStack is not running and the default test host app_db is not resolvable. CI/Paperclip should run the DB-backed tests in its environment.


ohld:

I found one blocking product-flow bug and one retry/idempotency bug. `/review` and the sensitive-surface `/cso` pass did not find secret leakage or SQL injection in the diff; `/codex review` independently flagged the retry/drop issue below.

[P1] src/flows/storage/memes.py:153 — Passed source polls process cached Telegram posts but never run or trigger the final meme pipeline. `process_cached_telegram_source()` ETLs and uploads the cached posts, but it does not call `final_meme_pipeline()`, `update_meme_status_of_ready_memes()`, or emit the same `ff.pipeline.telegram.completed` event that the normal Telegram pipeline uses to trigger finalization. Result: after moderators vote to add a source, the source is marked `parsing_enabled` and its media gets uploaded, but the new memes remain `created` with `telegram_file_id` and never become `ok`/recommendation-eligible until some unrelated final pipeline run happens. The feature's core promise is "vote passes -> source enters normal recommendations"; this path leaves it half-promoted.

Recommended fix: after `_process_unloaded_memes()` in the cached-source path, run the same finalization step used by the normal storage flow, or emit the existing completion event so Prefect automations run `final_meme_pipeline`. Add a test that closes a passed poll and verifies created cached memes become `ok` or that the finalization trigger is emitted.

[P2] src/storage/source_voting.py:496 — The daily selector only considers `status = 'discovered'` candidates with no existing `meme_source` and no prior poll. If `prepare_source_candidate()` succeeds but `send_message()`/`mark_source_candidate_poll_open()`/a retryable DB step fails before an open poll exists, the candidate is now `prepared` and/or has a matching `meme_source.url`, so later scheduled runs skip it forever. Codex independently flagged this as a persistent drop after transient failures.

Recommended fix: make the daily cycle resumable for prepared/cancelled/draft candidates, or roll back preparation/poll creation when posting fails. Add a regression test for "prepare succeeds, posting/opening fails, next cycle retries the same candidate".

Verification I ran:
- `ruff check` on changed implementation/test files: passed
- `python3 -m compileall` on changed implementation/test files: passed
- `pytest --collect-only` for new tests could not fully run in this host environment because project test dependencies are incomplete here (`bs4` missing); CI is the source of truth for DB-backed tests.

ohld:

I found three remaining blockers. `/review` structural pass and the sensitive-surface `/cso` check found no secret leakage or SQL injection, but `/codex review` independently found the same correctness risks below.

[P1] `src/flows/storage/memes.py:161` — `process_cached_telegram_source()` uploads cached memes and then calls `update_meme_status_of_ready_memes()` directly. That bypasses the normal `final_meme_pipeline()` checks at `src/flows/storage/memes.py:198-216`: caption/ad filtering, exact `telegram_file_id` duplicate resolution, and OCR duplicate resolution. A passed source vote can therefore make reposts or ad-like cached memes `ok` and recommendation-eligible before the normal final pipeline ever reviews them.

Recommended fix: route the cached-source path through `final_meme_pipeline()` or emit the existing finalization trigger instead of directly promoting ready memes. Add a regression test proving cached-source promotion runs duplicate/ad finalization before memes become `ok`.

[P2] `src/storage/source_voting.py:176` — `prepare_source_candidate()` creates or reuses a `meme_source` before scraping posts. If `fetch_telegram_candidate_posts()` or a later preparation step raises, the candidate remains `discovered` but now has a matching `meme_source.url`; `select_daily_source_candidate()` excludes discovered candidates with an existing source URL, so transient Telegram/scraper failures can strand the candidate forever.

Recommended fix: make preparation failures resumable, e.g. select discovered candidates linked to an `in_moderation` source with no poll, or roll back/mark the source/candidate consistently on preparation exceptions. Add a regression test for scrape failure followed by a later successful daily cycle.

[P2] `src/storage/source_voting.py:594` — `closes_at` is set when the draft poll is created, not when the message is actually opened. A draft retried after a failed Telegram send or a paused scheduler can be posted with `closes_at <= now`; `record_source_candidate_vote()` then rejects every vote as closed and the poll expires without a real voting window.

Recommended fix: set or refresh `closes_at = opened_at + SOURCE_VOTE_WINDOW` inside `mark_source_candidate_poll_open()`. Add a regression test for an old draft poll being opened with a fresh 24h window.

Verification I ran:
- `ruff check` on changed implementation/test files: passed
- `python3 -m compileall` on changed implementation/test files: passed
- GitHub CI: `lint` passed, `test` passed
- Local pytest collection could not run in this host environment because required settings env vars (`REDIS_URL`, `CORS_ORIGINS`, `CORS_HEADERS`) are absent.

ohld:
Post-merge note: this PR merged at 2026-05-14T16:01:03Z before I could cancel the queued merge. The changes-requested findings above are now post-merge production follow-up work, tracked in Paperclip as FFM-1260 assigned to CTO.

PR #258, opened 2026-05-15
Harden source-voting chat scope and poll targeting

ohld:
## Summary
- enforce moderator-chat-only voting in `record_source_candidate_vote` by validating both callback chat and poll chat against `TELEGRAM_MODERATOR_CHAT_ID`
- send new source-voting poll messages to `poll.chat_id` instead of a hardcoded chat id
- add regression tests for wrong-chat rejection and poll chat-id usage

## Verification
- `docker compose exec app pytest -q tests/storage/test_source_voting.py tests/storage/test_telegram_etl_source_status.py`
- result: `11 passed`

### DIFF
--- a/src/storage/source_voting.py
+++ b/src/storage/source_voting.py
@@ -316,6 +316,8 @@ async def record_source_candidate_vote(
     poll = await get_source_candidate_poll(poll_id)
     if poll is None:
         return {"status": "not_found"}
+    if chat_id != TELEGRAM_MODERATOR_CHAT_ID or poll["chat_id"] != TELEGRAM_MODERATOR_CHAT_ID:
+        return {"status": "wrong_chat", "poll": poll}
     if poll["chat_id"] != chat_id:
         return {"status": "wrong_chat", "poll": poll}
     if poll["status"] != POLL_STATUS_OPEN or poll["closes_at"] <= now:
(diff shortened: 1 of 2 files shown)


ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — The vote guard is moving in the right direction, but this still leaks the poll surface outside the moderator chat.

- [P2] `src/storage/source_voting.py:668` now sends resumed draft polls to `poll["chat_id"]`. If an existing draft poll row has a non-moderator `chat_id`, `post_new_source_candidate_poll`/`advance_daily_source_cycle` will publish the source-voting message into that non-moderator chat, while `record_source_candidate_vote` now rejects the exact same poll because `poll["chat_id"] != TELEGRAM_MODERATOR_CHAT_ID`. That creates an exposed, unvotable poll in the wrong chat. Cancel/reject non-moderator poll rows before sending, or constrain the send target to the moderator chat and keep the invariant single-sourced.

Security/CSO diff pass: no secrets or SQL-injection changes found, but this is a moderator chat-boundary regression.
Codex review: confirmed the same P2 issue.

ohld:
STAFF ENGINEER REVIEW: APPROVED — Review clean. Structural review, Codex review, and CSO diff pass found no actionable issues. The wrong-chat draft publish path is now cancelled before send, vote recording rejects non-moderator callback chats and poll rows, and no secrets/SQL injection changes were introduced.

ohld:
✅ Approved + merged.

PR #261, opened 2026-05-15
Refine source-voting poll text and finalize prepared-source dedup path

ohld:
## What
- shorten moderator poll message to compact Russian format
- auto-pin poll message silently (`disable_notification=true`)
- keep pinning best-effort (no poll failure on pin errors)
- in prepared-source activation path, run `final_meme_pipeline()` after upload so file_id/OCR dedup executes before memes become publishable
- update tests for silent pinning and prepared-source finalization path

## Why
- message was too long/noisy for moderator chat
- no NotifyAll requirement for pinning
- prepared-source cached ingest should use same dedup/finalization guarantees as regular pipeline

## Verification
- `pytest -q tests/flows/storage/test_cached_telegram_source.py`
- `python -m py_compile src/storage/source_voting.py src/flows/storage/memes.py tests/storage/test_source_voting.py tests/flows/storage/test_cached_telegram_source.py`
- DB-backed storage tests currently blocked locally by unresolved host `app_db` in this shell context

### DIFF
--- a/src/flows/storage/memes.py
+++ b/src/flows/storage/memes.py
@@ -159,7 +159,9 @@ async def process_cached_telegram_source(meme_source_id: int, limit: int = 100) 
         fresh_only=False,
     )
     await _process_unloaded_memes(unloaded_memes, "prepared Telegram source")
-    await update_meme_status_of_ready_memes()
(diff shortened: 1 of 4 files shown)


ohld:
Superseded by #262 (rebased cleanly on production).

PR #267, opened 2026-05-15
Rank inline search by meme quality

ohld:
## Summary
- return 20 inline search results instead of 10
- rank matched results by relevance, then meme quality/virality signals from meme_stats
- cover the result limit and SQL ordering with focused tests

## Ranking signals
- engagement_score
- smoothed like rate
- raw impression rank
- age
- reaction density
- invited/share count

## Verification
- python -m pytest tests/tgbot/test_inline_search.py -q
- python -m ruff check src/tgbot/service.py src/tgbot/handlers/inline.py tests/tgbot/test_inline_search.py
- python -m ruff format --check src/tgbot/service.py src/tgbot/handlers/inline.py tests/tgbot/test_inline_search.py
- python -m py_compile src/tgbot/service.py src/tgbot/handlers/inline.py tests/tgbot/test_inline_search.py
- git diff --check

### DIFF
--- a/src/tgbot/handlers/inline.py
+++ b/src/tgbot/handlers/inline.py
@@ -22,6 +22,7 @@ from src.tgbot.user_info import get_user_info
 
 MIN_SEARCH_QUERY_LENGTH = 3
 MAX_SEARCH_QUERY_LENGTH = 128
+INLINE_SEARCH_RESULT_LIMIT = 20
 INLINE_SEARCH_RESULT_CACHE_SECONDS = 60 * 60 * 12  # 12 hours
 
 
@@ -81,7 +82,7 @@ async def search_inline(update: Update, _: ContextTypes.DEFAULT_TYPE):
             ),
         )
 
-    memes = await search_memes_for_inline_query(query, limit=10)
+    memes = await search_memes_for_inline_query(query, limit=INLINE_SEARCH_RESULT_LIMIT)
 
     if len(memes) == 0:
         no_results_button = InlineQueryResultsButton(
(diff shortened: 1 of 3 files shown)


ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — two blockers before this can land.

1. The PR head drops the current production wildcard escaping in `search_memes_for_inline_query`. Production has `_escape_like_pattern()` and `ILIKE :search_pattern ESCAPE '\\'`; this branch builds `search_pattern` directly from user input again. Queries like `%%%` or `___` can broaden inline search to unrelated memes and make the expensive trigram/quality ranking scan far more than intended. Please rebase/merge current `production` and preserve the escaped LIKE pattern while adding the new ranking.

2. Codex review found that `COALESCE((MS.nlikes + 1.) / NULLIF(MS.nlikes + MS.ndislikes + 1, 0), 0.5)` ranks a `meme_stats` row with zero reactions as `1.0`, while a missing stats row is neutral `0.5`. `meme_stats` can be created by impression/share stats before reactions, so unvalidated memes can tie with actual 100% liked memes on equal text relevance. Use a neutral zero-reaction prior, for example `(nlikes + 1) / (nlikes + ndislikes + 2)` or an explicit zero-total `0.5` case.

Verification run:
- `python3 -m pytest tests/tgbot/test_inline_search.py -q` passed with test env overrides
- `python3 -m ruff check ...` passed
- `python3 -m ruff format --check ...` passed
- `git diff --check origin/production...HEAD` passed
- CI `lint` and `test` are green

PR #273, opened 2026-05-16
Fix fallback diagnostics reporting

ohld:
Fixes FFM-1296.

## Summary
- Record `diagnostics.fallback_used` only after a fallback engine returns candidates
- Add a regression test for the case where `text_light_lr_smoothed` is empty and `best_uploaded_memes` supplies the batch

## Tests
- `python3 -m pytest tests/recommendations/test_pipeline.py -q`

Note: local pytest needed explicit CI-style env vars and a missing host `sentry-sdk`/`pytest-env` install; the test passed after that environment setup.

### DIFF
--- a/src/recommendations/pipeline.py
+++ b/src/recommendations/pipeline.py
@@ -453,7 +453,6 @@ class RecommendationBatchPipeline:
         diagnostics: RecommendationBatchDiagnostics,
     ) -> list[Candidate]:
         for fallback in plan.fallback_engines:
-            diagnostics.fallback_used = diagnostics.fallback_used or fallback.engine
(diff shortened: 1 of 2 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — Clean structural review. The fallback diagnostics now record the engine that actually supplied candidates, and the regression test covers the multi-fallback empty-then-success path. Codex review found no issues. Local host pytest could not run because this runtime has an invalid/default SQLAlchemy dialect configuration, but GitHub CI lint and test are green.

ohld:
✅ Approved + merged.

PR #283, opened 2026-05-18
Shadow-log crossposting share signals

ohld:
## Summary
- add shadow-only in-bot share click fields to crossposting decision-log candidates
- keep live selection at score_version=2; no ranking formula, schema, or posting behavior change
- add reusable crossposting analytics SQL and a compact experiment readout with gate/rollback notes

## Hypothesis
RU memes with timestamp-safe pre-decision in-bot share clicks will better predict 24h Telegram channel forwards than aggregate likes alone. EN remains report-only until it shows a positive offline signal.

## Rollback
Revert this PR or ignore the new JSON fields in crossposting_decision_log.candidates. There is no migration, no score-version bump, and no changed selection behavior.

## Checks
- ruff check src/crossposting/service.py src/database.py tests/test_crossposting_meme.py
- ruff format --check src/crossposting/service.py src/database.py tests/test_crossposting_meme.py
- set -a; source ../ff-backend/.env.test; set +a; python3 -m pytest tests/test_crossposting_meme.py -q

## Production validation plan
After deploy health is green and not close to a scheduled channel slot:
1. Trigger Collect Channel Stats if stats freshness is stale.
2. Trigger one RU and one EN crossposting run.
3. Verify newest crossposting_decision_log rows contain pre_inbot_share_clicks and pre_inbot_share_click_users in candidates JSON.
4. Watch app/prefect logs and channel stats snapshots for the triggered posts.

PR #285, opened 2026-05-18
Fix: harden crossposting share ranker

ohld:
Fixes FFM-1327. Follow-up to #283 post-merge review findings.

## What changed
- Removed the unsafe `split_part(... )::bigint` parsing path from lateral share-count aggregation.
- Parse share deep links with a bounded canonical numeric regex capture so non-share/malformed/non-canonical rows cannot crash or skew the ranker.
- Added explicit `candidate_rank` ordering with a stable `M.id` tie-breaker and ordered the outer lateral-joined result before Python reads `rows[0]`.
- Added malformed/non-share deep link regression rows to the existing share-count test.
- Addressed Staff Engineer review by rejecting leading-zero sharer IDs in both RU and EN predicates and adding the `s_00010051_<meme_id>` self-click regression row.

## Verification
- `ruff format src/ tests/`
- `python3 -m py_compile src/crossposting/service.py tests/test_crossposting_meme.py`
- `DATABASE_URL="${DATABASE_URL/postgres:/postgresql+asyncpg:}" TELEGRAM_BOT_TOKEN="123456789:ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghi" python3 -m pytest tests/test_crossposting_meme.py -q` reached collection/execution: 9 caption tests passed; 8 DB-backed ranker tests errored because the host DB is not migrated (`relation "crossposting" does not exist`). Docker is unavailable in this runner, so the documented `docker compose exec app pytest ...` path could not be run locally.

## Staff Engineer notes
Please review the SQL ordering and canonical share-id parsing in `src/crossposting/service.py`. CI should run the integration tests against a migrated DB.

### DIFF
--- a/src/crossposting/service.py
+++ b/src/crossposting/service.py
@@ -219,7 +219,26 @@ _RU_QUERY = """
             MS.age_days, MS.nmemes_sent, MS.invited_count,
             SQ.signal AS src_signal,
             (SELECT m_signal FROM src_median) AS median_signal,
-            COUNT(*) OVER () AS candidate_pool_size
+            COUNT(*) OVER () AS candidate_pool_size,
+            ROW_NUMBER() OVER (
+                ORDER BY -1
+                    * COALESCE((MS.nlikes + 1.) / (MS.nlikes + MS.ndislikes + 1), 0.5)
+                    * CASE WHEN MS.raw_impr_rank <= 1 THEN 1 ELSE 0.8 END
+                    * CASE WHEN MS.age_days < 7 THEN 1 ELSE 0.8 END
+                    * CASE WHEN M.caption IS NULL THEN 1 ELSE 0.8 END
+                    * CASE
+                        WHEN MS.nmemes_sent <= 1 THEN 1
+                        ELSE (MS.nlikes + MS.ndislikes) * 1. / MS.nmemes_sent
+                      END
+                    * COALESCE(
+                        LEAST(2.0, GREATEST(0.5,
+                            SQ.signal / NULLIF((SELECT m_signal FROM src_median), 0)
+                        )),
+                        1.0
+                      )
+                    * (1.0 + LEAST(MS.invited_count, 10) * 0.1),
+                    M.id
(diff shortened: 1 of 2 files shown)


ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — `substring(... '^s_([0-9]{1,19})_' ...)` avoids the bigint crash, but it compares the captured sharer id as text. A malformed payload like `s_00010051_<meme_id>` still matches, and a self-click by user `10051` is counted as a non-self share because `10051 <> 00010051` as text. Please normalize the captured id before comparison or reject non-canonical numeric payloads in both RU and EN predicates, and add a regression row for the leading-zero self-click case.

ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — raw SQL/deep-link hardening still misses 19-digit overflow share IDs.

Finding:
- `src/crossposting/service.py:284` and `:393` still accept 19-digit overflow-sized sharer IDs such as `s_9999999999999999999_<meme_id>`. The regex captures them, the text comparison treats them as a different non-self user, and malformed deep links can still skew `pre_inbot_share_clicks` / `pre_inbot_share_click_users`. The added regression covers a 21-digit value, but not the max-length overflow case.

Fix: reject or safely normalize only sharer IDs valid for the `user_id` bigint domain before comparing, and add a 19-digit overflow regression covering the ranker query path.

CSO/security pass: no hardcoded secrets in the PR diff; no new user-controlled SQL interpolation beyond `ranked.id` and parameterized `:limit`.

Verification: clean worktree diff is 2 files. Local targeted pytest is blocked in this runtime because `app_db` does not resolve; GitHub CI remains the merge gate. Auto-merge is disabled.

ohld:
STAFF ENGINEER REVIEW: APPROVED — /review, /codex review, and raw-SQL security pass found no blocking issues. CI lint/test are green; share-link parsing is bounded/canonical and candidate ordering is deterministic.

ohld:
✅ Approved + merged.

PR #290, opened 2026-05-20
Fix: disable unregistered text-light blender exposure

ohld:
Fixes FFM-1357.

## What changed
- Disables the `text_light_blender_v1` overlay in the production queue generation path.
- Flips the `RecommendationBatchRequest` default to disabled so callers must opt in explicitly.
- Updates the queue regression test to assert the text-light assignment helper is not called by default.
- Updates the recommendation spec to reflect that text-light overlay requires a registered experiment before exposure.

## Root cause
The `generate_recommendations()` hot path hard-coded `text_light_blender_v1_enabled=True`, so queue refills could create and honor `text_light_blender_v1` assignments even though no CEO-owned active experiment record existed.

## Verification
- `ruff format src/ tests/`
- `python3 -m pytest tests/recommendations/test_meme_queue.py tests/recommendations/test_pipeline.py -q`

### DIFF
--- a/specs/recommendations.md
+++ b/specs/recommendations.md
@@ -38,11 +38,11 @@ Removed engines: `fast_dopamine`, `classic`, `multiply_all_scores`, `selected_so
 | Stage | Trigger | Engines | Source |
 |-------|---------|---------|--------|
 | Cold start | nmemes_sent < 30 | 3-phase adaptive with text-light guards: explore/adapt/fallback avoid OCR text above 30 words | [`pipeline.py`](../src/recommendations/pipeline.py) |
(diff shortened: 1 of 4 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — structural review clean. /review found no SQL/data safety, race, shell, LLM trust-boundary, enum-completeness, or recommendation weight invariant issues. /codex review found no actionable findings. CI is green on the PR head. Local targeted pytest could not collect in this runtime because SQLAlchemy could not load the postgres dialect under Python 3.13, so GitHub CI is the test authority here.

ohld:
✅ Approved + merged.

PR #318, opened 2026-06-17
Fix: send first meme during onboarding

ohld:
Fixes FFM-1606.

## What changed
- Created non-share /start users now continue from language settings into the existing onboarding_flow, which sends the first meme via next_message.
- Share-link onboarding, kitchen/wrapped utility links, blocked acquisition channels, and existing-user paths keep their prior routing.
- Added orchestration coverage for the no-DB created-user path and updated start-handler regression expectations.

## Root cause
The created-user /start branch persisted user/user_tg rows and returned after handle_language_settings for plain starts and giveaway starts. That left a user row without a user_meme_reaction delivery row unless the user later pressed the language-settings completion button.

## Verification
- ruff check --fix src/ tests/
- ruff format src/ tests/
- pytest tests/tgbot/test_start.py::test_new_user_no_deep_link_enters_onboarding_without_db -q
- pytest tests/tgbot/test_start.py --collect-only -q

## Not run
- Full DB-backed tests: local Postgres was not running on localhost:5432 and docker is not installed in this environment.

### DIFF
--- a/src/tgbot/handlers/language.py
+++ b/src/tgbot/handlers/language.py
@@ -9,6 +9,7 @@ from src.recommendations.meme_queue import (
     clear_meme_queue_for_user,
     generate_recommendations,
 )
+from src.recommendations.service import get_user_reactions
 from src.tgbot.constants import LANG_SETTINGS_END_CALLBACK_DATA
 from src.tgbot.handlers.onboarding import onboarding_flow
 from src.tgbot.service import (
@@ -138,5 +139,8 @@ async def handle_language_settings_end(
     user_id = update.effective_user.id
     await clear_meme_queue_for_user(user_id)
     await generate_recommendations(user_id, limit=15)
+
+    if await get_user_reactions(user_id):
+        return None
 
     return await onboarding_flow(update, context.bot)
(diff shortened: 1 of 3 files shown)


ohld:
STAFF ENGINEER REVIEW: CHANGES REQUESTED — Two production-visible onboarding regressions found; requesting changes.

1. New `giveaway_*` users now get duplicate meme delivery. `handle_giveaway()` already calls `next_message()` for both known and unknown campaigns, then the new fall-through calls `onboarding_flow()`, which calls `next_message()` again. The existing-user giveaway path still returns after `handle_giveaway()`, so this only regresses newly-created users.

2. Plain new users can replay onboarding by pressing the language settings Done button. `handle_language_settings()` still renders the existing Done callback, and `handle_language_settings_end()` still calls `onboarding_flow()`. Since PR #318 now runs onboarding immediately after rendering the prompt, pressing Done later sends the welcome/countdown and another meme again.

Please make the new immediate-first-meme path idempotent with the existing language completion and giveaway paths, and add regression coverage for both duplicate-delivery cases.

ohld:
STAFF ENGINEER REVIEW: APPROVED — Clean on refreshed head be9a3c2. Structural review + Codex review found the prior duplicate-delivery issues fixed: giveaway starts now return after handle_giveaway, and language Done no longer replays onboarding after a delivery row exists. Secret/raw-SQL diff scan clean.

ohld:
✅ Approved + merged.

PR #320, opened 2026-06-18
Fix E2E smoke URL buttons

ohld:
Fixes FFM-1629.

## What changed
- Added a safe helper for reading Telethon callback button data so URL inline buttons without `.data` are treated as non-callback buttons.
- Reused the helper in reaction-button detection, like-click selection, and `/delete` confirmation scanning.
- Added regression coverage for URL-only and URL-before-reaction keyboards.

## Root cause
The smoke script scanned every inline keyboard button using `btn.data`. Telethon URL buttons expose `.url` instead of `.data`, so a share URL button could crash the smoke before it reached the reaction buttons.

## Verification
- `pytest tests/scripts/test_e2e_smoke.py`
- `ruff check --fix scripts/e2e_smoke.py tests/scripts/test_e2e_smoke.py`
- `ruff format scripts/e2e_smoke.py tests/scripts/test_e2e_smoke.py`

### DIFF
--- a/scripts/e2e_smoke.py
+++ b/scripts/e2e_smoke.py
@@ -17,6 +17,7 @@ Usage:
     python scripts/e2e_smoke.py           # returning user flow
     python scripts/e2e_smoke.py --fresh   # reset state + onboarding flow
 """
+
 import argparse
 import asyncio
 import os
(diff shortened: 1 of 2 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — Clean structural review. Scope is limited to E2E smoke callback-button handling; focused regression tests, local pytest/Ruff checks, GitHub CI, and Codex review are clean.

ohld:
✅ Approved + merged.

PR #326, opened 2026-06-27
Fix: suppress handled chat agent max-turn Sentry events

ohld:
Fixes FFM-1764.

## Summary
- Add chat-agent Sentry scope around the OpenAI Agents runner call so agent events carry explicit chat-agent context.
- Drop expected `agents.exceptions.MaxTurnsExceeded` events from the chat-agent path even when Sentry reports them with `logger=null` / `mechanism=openai_agents`.
- Add regression coverage for both scoped chat-agent events and the production webhook event shape.

## Verification
- `python3 -m pytest tests/test_sentry_observability.py -q`

## Notes
- Host Python lacks the app `agents.exceptions` dependency and Docker is unavailable here, so `tests/tgbot/test_chat_agent_runner.py` could not be collected locally. CI should run it in the app dependency environment.

### DIFF
--- a/src/observability/sentry.py
+++ b/src/observability/sentry.py
@@ -64,6 +64,33 @@ def before_send_log(log: dict[str, Any], _hint: dict[str, Any]) -> dict[str, Any
 
 
 @contextmanager
+def chat_agent_scope(
+    *,
+    chat_id: int,
+    user_id: int,
+    reply_to_message_id: int | None = None,
+    trigger_type: str | None = None,
+) -> Iterator[None]:
+    """Attach chat-agent context to Sentry events produced while running the agent."""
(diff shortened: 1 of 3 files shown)


ohld:
STAFF ENGINEER REVIEW: APPROVED — clean structural review. Checked Sentry suppression scope, chat-agent context lifetime, Agents SDK call sites, secret patterns, and targeted regressions. Codex review found no discrete regression. Local verification: `python3 -m pytest tests/test_sentry_observability.py -q` passed; `tests/tgbot/test_chat_agent_runner.py` could not collect on host because `agents.exceptions` is unavailable in this runtime.

ohld:
✅ Approved + merged.

